PR #5, opened 2017-09-04
Clean up DAO

rkondratowicz:
Cleaning up a bit.

```
[ ] Yes
[x] No
```


rkondratowicz:
I'm tempted to just inline those sql queries (all used only once), what do you guys think?

PR #9, opened 2017-09-06
Read all drafts for user

rkondratowicz:
### Change description ###

Retrieve all drafts for given user and type.
The old endpoint uses new repo method as well, but still returns data in old format.

```
[ ] Yes
[x] No
```


src/functionalTest/java/uk/gov/hmcts/reform/draftstore/data/DraftStoreDAOTest.java
```diff
@@ -97,22 +99,20 @@ public void shouldRetrieve() throws SQLException {
...
-        assertThat(document).isEqualTo("{ \"test\":\"1234\"}");
+        assertThat(drafts.size()).isEqualTo(1);
+        assertThat(drafts.get(0).document).isEqualTo("{ \"test\":\"1234\"}");
         assertNoOtherUserDataHasBeenAffected();
     }
 
-    @Test(expected = NoDraftFoundException.class)
-    public void shouldThrowOnRetrieveWhenNoDocumentForUser() throws SQLException {
-        underTest.retrieve(USER_ID, "default");
-    }
```
rkondratowicz: repo no longer throw exception, moved to controller

lgonczar: I'd replace that test with one that checks if the DAO returns an empty list.

lgonczar: It's also worth adding a test checking if we can retrieve multiple drafts.

src/main/java/uk/gov/hmcts/reform/draftstore/data/DraftStoreDAO.java
```diff
@@ -62,15 +62,14 @@ public SaveStatus insertOrUpdate(String userId, String type, String newDocument)
...
-    public String retrieve(String userId, String type) {
-        MapSqlParameterSource params =
+    public List<Draft> readAll(String userId, String type) {
+        return jdbcTemplate.query(
+            "SELECT * FROM draft_document WHERE user_id = :userId AND document_type = :type",
             new MapSqlParameterSource()
                 .addValue("userId", userId)
-                .addValue("type", type);
-
-        List<String> documents = jdbcTemplate.query(QUERY, params, (rs, rowNum) -> rs.getString(1));
```
lgonczar: QUERY is probably no longer used

rkondratowicz: removed

PR #22, opened 2017-09-14
Change draft id type to string

rkondratowicz:
## Change description ###
What: `id` field on returned model is now a string, not an int
Why: https://hmcts.github.io/restful-api-standards/#common-data-types

```
[x] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/reform/draftstore/endpoint/v3/DraftController.java
```diff
@@ -157,4 +157,12 @@ private void assertCanEdit(Draft draft, String userId) {
             throw new AuthorizationException();
         }
     }
+
+    private int toDbId(String apiId) {
```
njrich28: Maybe document this method to explain the situation for people looking at this in the future

PR #27, opened 2017-09-15
Deprecate v2 api

rkondratowicz:

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/draftstore/endpoint/v2/DraftStoreEndpoint.java
+++ b/src/main/java/uk/gov/hmcts/reform/draftstore/endpoint/v2/DraftStoreEndpoint.java
@@ -11,6 +11,7 @@ import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestHeader;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
+import uk.gov.hmcts.reform.api.deprecated.APIDeprecated;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/reform/draftstore/endpoint/v2/DraftStoreEndpoint.java
```diff
@@ -27,6 +28,12 @@
 import static org.springframework.web.bind.annotation.RequestMethod.POST;
 import static uk.gov.hmcts.reform.draftstore.domain.SaveStatus.Updated;
 
+@APIDeprecated(
+    name = "/api/v2/draft endpoint",
+    expiryDate = "TBD",
+    docLink = "about:blank",
+    note = "Use new `/drafts` endpoint instead."
+)
```
rkondratowicz: Can we make expiryDate and docLink optional? If now, are those values ok?

njrich28: What is the expiry date? i.e. when the api will be removed?

rkondratowicz: well, we don't know

njrich28: would probably be best to leave it blank if it's not defined and have it conform to ISO 8601 format so that it can be processed by machine

rkondratowicz:
Closing for now as expiry date is not known yet

PR #30, opened 2017-09-19
use idam for user auth

rkondratowicz:
Handle jwt in `Authorization` header (users only for now) in the new api by sending a request to Idam.

```
[x] Yes (v3 only)
[ ] No
```

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -37,6 +37,10 @@ info:
   app:
     name: ${spring.application.name}
 
+idam:
+  url: ${IDAM_URL}
+  useStub: ${IDAM_USE_STUB:false}
+
 ---
 spring:
   profiles: test
@@ -50,3 +54,5 @@ spring:
   datasource:
     url: jdbc:postgresql://localhost:${DRAFT_STORE_DB_PORT:5432}/draftstore
     password: ${DRAFT_STORE_DB_PASSWORD:}
+idam:
+  useStub: true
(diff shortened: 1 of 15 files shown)


src/main/resources/application.yaml
```diff
@@ -50,3 +54,5 @@ spring:
   datasource:
     url: jdbc:postgresql://localhost:${DRAFT_STORE_DB_PORT:5432}/draftstore
     password: ${DRAFT_STORE_DB_PASSWORD:}
+idam:
+  useStub: true
```
timja-kainos: shouldn't you use the real one in functionalTests to make sure it works properly?

src/main/resources/application.yaml
```diff
@@ -37,6 +37,10 @@ info:
   app:
     name: ${spring.application.name}
 
+idam:
+  url: ${IDAM_URL}
+  useStub: ${IDAM_USE_STUB:false}
```
timja-kainos: is the stub needed?

src/main/resources/application.yaml
rkondratowicz: This profile is used in only one weird test (`HandleUnknownExceptionTest`) where repo is actually mocked and set up to throw an exception. Not really a functional test.

I'll look into it on a separate occasion.

src/main/resources/application.yaml
rkondratowicz: only for people who want to run the app without having access to idam.

lgonczar: Worth considering making an external stub, so that the draft store would unaware of stubbing. Without that stub being present, I'm ok with the code as it is.

PR #31, opened 2017-09-20
use s2s

rkondratowicz:
Retrieve name of the service calling draft store by passing the header to s2s api.

```
[x] Yes (in v3 only)
[ ] No
```

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -42,7 +42,7 @@ idam:
   useStub: ${IDAM_USE_STUB:false}
 
 s2s:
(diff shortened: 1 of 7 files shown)


src/main/resources/application.yaml
```diff
@@ -41,6 +41,10 @@ idam:
   url: ${IDAM_URL}
   useStub: ${IDAM_USE_STUB:false}
 
+s2s:
+  url: ${S2S_URL}
```
timja-kainos: add default?

rkondratowicz: done

PR #32, opened 2017-09-20
Add `application-default.yaml` to gitignore...

rkondratowicz:
### Change description ###
... so that you can configure the app by setting up an untracked file instead of environment variables.

https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html#boot-features-external-config-profile-specific-properties

```
[ ] Yes
[x] No
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ build
 .DS_store
 .gradle
 out/
+
+application-default.yaml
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -3,3 +3,5 @@ build
 .DS_store
 .gradle
 out/
+
+application-default.yaml
```
timja-kainos: add to readme maybe?

rkondratowicz: done

PR #34, opened 2017-09-21
Add `created` and `updated` fields

rkondratowicz:
### Change description ###
Store information on when draft has be created and last updated.
Using https://en.wikipedia.org/wiki/ISO_8601 format.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/functionalTest/java/uk/gov/hmcts/reform/draftstore/data/DataAgent.java
+++ b/src/functionalTest/java/uk/gov/hmcts/reform/draftstore/data/DataAgent.java
@@ -27,8 +27,8 @@ public class DataAgent {
         jsonbObj.setValue(document);
 
         jdbcTemplate.update(
-            "INSERT INTO draft_document (user_id, document_type, service, document) "
-                + "VALUES (:userId, :type, 'cmc', :document)",
+            "INSERT INTO draft_document (user_id, document_type, service, document, created, updated) "
+                + "VALUES (:userId, :type, 'cmc', :document, now(), now())",
             new MapSqlParameterSource("userId", userId)
                 .addValue("type", type)
                 .addValue("document", jsonbObj)
--- a/src/functionalTest/java/uk/gov/hmcts/reform/draftstore/endpoint/v3/helpers/SampleData.java
+++ b/src/functionalTest/java/uk/gov/hmcts/reform/draftstore/endpoint/v3/helpers/SampleData.java
@@ -1,0 +1,19 @@
+package uk.gov.hmcts.reform.draftstore.endpoint.v3.helpers;
+
+import uk.gov.hmcts.reform.draftstore.domain.Draft;
+
+import java.time.LocalDateTime;
+
+public class SampleData {
+    public static Draft draft(String id) {
+        return new Draft(
+            id,
+            "abc",
+            "serviceA",
+            "",
+            "",
+            LocalDateTime.now(),
+            LocalDateTime.now()
+        );
+    }
+}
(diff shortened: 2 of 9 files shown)


src/functionalTest/java/uk/gov/hmcts/reform/draftstore/endpoint/v3/helpers/SampleData.java
```diff
@@ -0,0 +1,19 @@
...
+
+import java.time.ZonedDateTime;
+
+public class SampleData {
+    public static Draft draft(String id) {
+        return new Draft(
+            id,
+            "abc",
+            "serviceA",
+            "",
```
lgonczar: Can these values normally be empty?

src/main/java/uk/gov/hmcts/reform/draftstore/data/DraftStoreDAO.java
```diff
@@ -47,12 +49,15 @@ public DraftStoreDAO(NamedParameterJdbcTemplate jdbcTemplate) {
...
     public SaveStatus insertOrUpdate(String userId, String service, String type, String newDocument) {
+        ZonedDateTime now = ZonedDateTime.now();
         MapSqlParameterSource params =
             new MapSqlParameterSource()
                 .addValue("userId", userId)
                 .addValue("service", service)
                 .addValue("type", type)
-                .addValue("document", newDocument);
+                .addValue("document", newDocument)
+                .addValue("created", now)
```
lgonczar: Can this overwrite `created` on update?

src/main/java/uk/gov/hmcts/reform/draftstore/data/DraftStoreDAO.java
```diff
@@ -156,7 +167,9 @@ public Draft mapRow(ResultSet rs, int rowNumber) throws SQLException {
                 rs.getString("user_id"),
                 rs.getString("service"),
                 rs.getString("document"),
-                rs.getString("document_type")
+                rs.getString("document_type"),
+                rs.getTimestamp("created").toInstant().atZone(ZoneId.systemDefault()),
```
lgonczar: We may consider extracting a conversion method. Not that important for now, as there are just two lines like this.

src/main/java/uk/gov/hmcts/reform/draftstore/data/DraftStoreDAO.java
rkondratowicz: it's not used in update query

src/main/resources/db/migration/V5__add_created_and_updated_columns.sql
```diff
@@ -0,0 +1,19 @@
+ALTER TABLE draft_document
```
lgonczar: We could do these two things (created, updated) in one go, but only if migration time matters here

src/functionalTest/java/uk/gov/hmcts/reform/draftstore/data/DataAgent.java
```diff
@@ -27,8 +27,8 @@ public void setupDocumentForUser(String userId, String type, String document) th
         jsonbObj.setValue(document);
 
         jdbcTemplate.update(
-            "INSERT INTO draft_document (user_id, document_type, service, document) "
-                + "VALUES (:userId, :type, 'cmc', :document)",
+            "INSERT INTO draft_document (user_id, document_type, service, document, created, updated) "
+                + "VALUES (:userId, :type, 'cmc', :document, now(), now())",
```
timja-kainos: just putting it out there, what we do in the claim store is always insert at UTC and then use in LocalDateTime which is easier to work with.

rkondratowicz: We return timezone info (although it's always UTC) to follow https://hmcts.github.io/restful-api-standards/#126

PR #35, opened 2017-09-22
Add max stale time, remove old drafts periodically

rkondratowicz:
### Change description ###
Drafts created via old api (v2) do not get removed.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/functionalTest/java/uk/gov/hmcts/reform/draftstore/data/OldDraftsCleanupTest.java
+++ b/src/functionalTest/java/uk/gov/hmcts/reform/draftstore/data/OldDraftsCleanupTest.java
@@ -1,0 +1,79 @@
+package uk.gov.hmcts.reform.draftstore.data;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.test.context.ActiveProfiles;
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+import uk.gov.hmcts.reform.draftstore.endpoint.v3.helpers.SampleData;
+
+import java.time.Clock;
+import java.time.Duration;
+import java.time.Instant;
+import java.time.ZoneId;
+import java.util.function.Consumer;
+
+import static java.time.Instant.now;
+import static java.time.temporal.ChronoUnit.DAYS;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringBootTest
+@ActiveProfiles("test")
+public class OldDraftsCleanupTest {
+
+    @Autowired
+    private NamedParameterJdbcTemplate jdbcTemplate;
+    private Instant now = now();
+
+    @Test
+    public void should_remove_stale_drafts() throws Exception {
+        DraftStoreDAO nowRepo = repoAtTime(now);
+
+        int d1 = nowRepo.insert("", "", SampleData.createDraft(5));
+        int d2 = nowRepo.insert("", "", SampleData.createDraft(15));
+
+        after(
+            Duration.of(10, DAYS),
+            repo -> {
+                repo.deleteStaleDrafts();
+                assertThat(repo.read(d1)).isEmpty();
+                assertThat(repo.read(d2)).isNotEmpty();
+            }
+        );
+    }
+
+    @Test
+    public void should_NOT_remove_old_drafts_that_were_recently_updated() throws Exception {
+
+        int newDraftId = repoAtTime(now).insert("", "", SampleData.createDraft(10));
+
+        after(
+            Duration.of(9, DAYS),
+            repo -> repo.update(newDraftId, SampleData.updateDraft())
+        );
+
+        after(
+            Duration.of(11, DAYS),
+            repo -> {
+                repo.deleteStaleDrafts();
+                assertThat(repo.read(newDraftId)).isNotEmpty();
+            }
+        );
+    }
+
+    private DraftStoreDAO repoAtTime(Instant instant) {
+        return new DraftStoreDAO(
+            jdbcTemplate,
+            0,
+            Clock.fixed(instant, ZoneId.systemDefault())
+        );
+    }
+
+    private void after(Duration duration, Consumer<DraftStoreDAO> daoOperation) {
+        DraftStoreDAO dao = repoAtTime(now.plus(duration));
+        daoOperation.accept(dao);
+    }
+}
--- a/src/main/java/uk/gov/hmcts/reform/draftstore/DraftStoreApplication.java
+++ b/src/main/java/uk/gov/hmcts/reform/draftstore/DraftStoreApplication.java
@@ -2,8 +2,10 @@ package uk.gov.hmcts.reform.draftstore;
 
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.scheduling.annotation.EnableScheduling;
 
 @SpringBootApplication
+@EnableScheduling
 public class DraftStoreApplication {
 
     public static final String BASE_PACKAGE_NAME = DraftStoreApplication.class.getPackage().getName();
--- a/src/main/java/uk/gov/hmcts/reform/draftstore/config/DraftStoreConfig.java
+++ b/src/main/java/uk/gov/hmcts/reform/draftstore/config/DraftStoreConfig.java
@@ -46,7 +46,7 @@ public class DraftStoreConfig {
         return new DraftStoreDAO(
             jdbcTemplate,
             maxStaleDaysDefault,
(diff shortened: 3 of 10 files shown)


src/main/java/uk/gov/hmcts/reform/draftstore/config/DraftStoreConfig.java
```diff
@@ -28,6 +28,7 @@
 
     @Value("${idam.url}") private String idamUrl;
     @Value("${s2s.url}") private String s2sUrl;
+    @Value("${maxAge.default}") private int maxAgeDefault;
```
lgonczar: Regarding naming `maxAgeDefault` or `maxAge`, I'm wondering whether it would be sensible to add some information about the unit, i.e. `maxAgeInDays` or something. What do you think?

yarpo: Why not `storeForNoOfDays`?

rkondratowicz: I'll see if Spring can accept something like `"2 days"` which then maps to Java's `Duration`

src/main/java/uk/gov/hmcts/reform/draftstore/DraftStoreApplication.java
```diff
@@ -2,8 +2,10 @@
 
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.scheduling.annotation.EnableScheduling;
 
 @SpringBootApplication
+@EnableScheduling
```
timja-kainos: Looking at the current setup it appears that this scheduled job would start on all nodes?
since its deleting it shouldn't be a problem? but worth calling out

src/main/java/uk/gov/hmcts/reform/draftstore/DraftStoreApplication.java
```diff
@@ -2,8 +2,10 @@
...
 public class DraftStoreApplication {
```
timja-kainos: also I thought there was a need for a job scheduling component in reform ;)? 
We decided we aren't doing this?

src/main/java/uk/gov/hmcts/reform/draftstore/DraftStoreApplication.java
rkondratowicz: Yeah, no danger of doing something more than once in this case :)

src/main/resources/application.yaml
```diff
@@ -45,6 +45,10 @@ s2s:
   url: ${S2S_URL:}
   useStub: ${S2S_USER_STUB:false}
 
+maxAge:
+  default: ${MAX_AGE_DEFAULT:14}
```
timja-kainos: is there any reason for this default?
I would have thought at least a month, probably longer, merging this will get deployed in our draft store we deploy so need to think about either the default or we just set it for a long time

src/main/java/uk/gov/hmcts/reform/draftstore/data/DraftStoreDAO.java
```diff
@@ -160,6 +163,14 @@ public void delete(int id) {
         );
     }
 
+    public void deleteOldDrafts() {
+        jdbcTemplate.update(
+            "DELETE FROM draft_document "
+                + "WHERE updated + interval '1 day' * max_age < now()",
```
lgonczar: Is there an index on `updated`? If not, we could add it. Not sure if it matters (because the DBMS might be smart about those things), but maybe index-wise it would be safer to move '1 day' to the other side, i.e. 
```
    WHERE updated max_age < now() - interval '1 day'
```

There's also a question of whether we want to delete drafts last updated 'up until 24th August 2017' or 'up until 24th August 2017 12:34:56.789'. Maybe rounding to midnight would give us some benefits, for example if the query fails, we know it failed for all drafts from a given date. Again, not sure if it matters to us at all.

src/main/resources/application.yaml
rkondratowicz: This change does not affect old api, drafts are safe.

The default max age value is open for discussion :)

src/main/java/uk/gov/hmcts/reform/draftstore/service/ScheduledTasks.java
```diff
@@ -0,0 +1,20 @@
...
+@Component
+public class ScheduledTasks {
+
+    private final DraftStoreDAO repo;
+
+    public ScheduledTasks(DraftStoreDAO repo) {
+        this.repo = repo;
+    }
+
+    @Scheduled(cron = "${maxAge.cron}")
```
lgonczar: How about `deleteOldDrafts.cron`?

src/main/java/uk/gov/hmcts/reform/draftstore/config/DraftStoreConfig.java
rkondratowicz: just realised that `maxAge` is not the best name as it suggests it's time from when draft was created, not when it was last updated...  
Perhaps `maxStaleDays`? What do you thinks? Also, github needs polls ;)

src/main/java/uk/gov/hmcts/reform/draftstore/service/ScheduledTasks.java
rkondratowicz: it was `cron.defaultOldDrafts` initially, but I wanted to group default max age and cron in config :)

src/main/java/uk/gov/hmcts/reform/draftstore/data/DraftStoreDAO.java
KarolJastrzebski: Replace `now()` with a parameter to enable time-based unit testing.

yarpo:
```
now() at time zone 'utc'
```
?

src/main/java/uk/gov/hmcts/reform/draftstore/config/DraftStoreConfig.java
rlewan: `maxStaleDays` sounds good to me.

src/main/java/uk/gov/hmcts/reform/draftstore/data/DraftStoreDAO.java
lgonczar: I think the DB should be configured to default to UTC instead.

src/main/java/uk/gov/hmcts/reform/draftstore/config/DraftStoreConfig.java
```diff
@@ -42,7 +43,11 @@ public Filter etagFilter() {
 
     @Bean
     public DraftStoreDAO draftDocumentDAO(NamedParameterJdbcTemplate jdbcTemplate) {
-        return new DraftStoreDAO(jdbcTemplate, maxStaleDaysDefault);
+        return new DraftStoreDAO(
+            jdbcTemplate,
+            maxStaleDaysDefault,
+            Clock.systemUTC()
```
lgonczar: I'd let the time zone settings be configured at system property level.

src/functionalTest/java/uk/gov/hmcts/reform/draftstore/data/OldDraftsCleanupTest.java
```diff
@@ -0,0 +1,79 @@
...
+    @Autowired
+    private NamedParameterJdbcTemplate jdbcTemplate;
+    private Instant now = now();
+
+    @Test
+    public void should_remove_stale_drafts() throws Exception {
+        DraftStoreDAO nowRepo = repoAtTime(now);
+
+        int d1 = nowRepo.insert("", "", SampleData.createDraft(5));
+        int d2 = nowRepo.insert("", "", SampleData.createDraft(15));
```
lgonczar: Not that important, but those values could be 10 and 11 - this way we would have the behaviour around the threshold fully checked.

src/functionalTest/java/uk/gov/hmcts/reform/draftstore/data/OldDraftsCleanupTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+        after(
```
lgonczar: nice

src/main/java/uk/gov/hmcts/reform/draftstore/config/DraftStoreConfig.java
```diff
@@ -41,7 +43,11 @@ public Filter etagFilter() {
 
     @Bean
     public DraftStoreDAO draftDocumentDAO(NamedParameterJdbcTemplate jdbcTemplate) {
-        return new DraftStoreDAO(jdbcTemplate);
+        return new DraftStoreDAO(
+            jdbcTemplate,
+            maxStaleDaysDefault,
+            Clock.systemUTC()
```
lgonczar: I'd let the time zone settings be configured at system property level (is there a way not to provide it here?). This way we would always have just one point where time zone is set.

PR #36, opened 2017-09-25
Enable PMD gradle plugin

KarolJastrzebski:
This is part of quality assurance activity to standardise the code. Few simple rules (basic, braces, unnecessary and unusedcode) should be enough for a good start.

```
[ ] Yes
[x] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -117,7 +117,7 @@ sonarqube {
 pmd {
   toolVersion = "5.8.1"
   ignoreFailures = true
-  sourceSets = [sourceSets.main]
+  sourceSets = [sourceSets.main, sourceSets.test, sourceSets.functionalTest]
   reportsDir = file("$project.buildDir/reports/pmd")
   ruleSetFiles = files("ruleset.xml")
 }
--- a/build.gradle
+++ b/build.gradle
@@ -119,12 +119,7 @@ pmd {
   ignoreFailures = true
   sourceSets = [sourceSets.main]
   reportsDir = file("$project.buildDir/reports/pmd")
-  ruleSets = [
-    "java-basic",
-    "java-braces",
-    "java-unnecessary",
-    "java-unusedcode"
-  ]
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -113,6 +114,19 @@ sonarqube {
   }
 }
 
+pmd {
+  toolVersion = "5.8.1"
+  ignoreFailures = true
+  sourceSets = [sourceSets.main]
```
rkondratowicz: add `test` and `functionalTest`

build.gradle
```diff
@@ -113,6 +114,19 @@ sonarqube {
   }
 }
 
+pmd {
```
timja-kainos: Is this instead of checkstyle?
i.e. this repo already has checkstyle on it

KarolJastrzebski: Rather a complementary check to be used within Sonar or Codacy.

build.gradle
KarolJastrzebski: `pmd` for all sources is run during `check` task, and `pmdFunctionalTests` is also available.

rkondratowicz: This property is used to configure what source sets should be analyzed as part of the `check` and `build` tasks.

build.gradle
timja-kainos: What benefit does it give? How much extra time does it take?

ruleset.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0"?>
+<ruleset name="Custom ruleset"
```
rkondratowicz: perhaps 'Draft Store' ruleset or something similar?

build.gradle
KarolJastrzebski: It helps finding bad practices and takes as much as caring for quality does. ;)

timja-kainos: shouldn't sonar be used for this?

PR #38, opened 2017-09-26
Filter drafts by json doc content

rkondratowicz:
### Change description ###

Allows simple filtering by json document content using query params.  
For example:
```
/drafts?colour=red&size=large
```
Only string properties are supported.  

There was an endpoint previously that allowed filtering only by type (`/drafts?type=xxx`) which checked special 'type' column, it is no longer supported.    
The same functionality can be now achieved by adding 'type' field to actual json document.

```
[x] Yes (v3 only)
[ ] No
```

### DIFF
--- a/src/functionalTest/java/uk/gov/hmcts/reform/draftstore/data/OldDraftsCleanupTest.java
+++ b/src/functionalTest/java/uk/gov/hmcts/reform/draftstore/data/OldDraftsCleanupTest.java
@@ -1,5 +1,6 @@
 package uk.gov.hmcts.reform.draftstore.data;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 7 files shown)


src/functionalTest/java/uk/gov/hmcts/reform/draftstore/data/OldDraftsCleanupTest.java
```diff
@@ -24,8 +25,8 @@
 @ActiveProfiles("test")
 public class OldDraftsCleanupTest {
 
-    @Autowired
-    private NamedParameterJdbcTemplate jdbcTemplate;
+    @Autowired private NamedParameterJdbcTemplate jdbcTemplate;
```
timja-kainos: why move these lines?

src/functionalTest/java/uk/gov/hmcts/reform/draftstore/data/OldDraftsCleanupTest.java
```diff
@@ -24,8 +25,8 @@
...
+    @Autowired private ObjectMapper objectMapper;
```
timja-kainos: and this?

PR #45, opened 2017-10-03
Make it possible for apps to start using docker-compose

lgonczar:
### Change description ###

Make it possible for apps to start using docker-compose.
Before this fix, database's healthcheck wouldn't work.

```
[ ] Yes
[x] No
```


docker-compose.yml
```diff
@@ -12,13 +12,16 @@ services:
...
+      - DRAFT_STORE_DB_PORT
       - DRAFT_STORE_DB_PASSWORD
       - ROOT_APPENDER
       - JSON_CONSOLE_PRETTY_PRINT
       - ROOT_LOGGING_LEVEL
       - REFORM_SERVICE_NAME
       - REFORM_TEAM
       - REFORM_ENVIRONMENT
+      - IDAM_USE_STUB
+      - S2S_USER_STUB
```
rkondratowicz: typo

docker/database/Dockerfile
```diff
@@ -2,6 +2,6 @@ FROM postgres:9.6
 
 COPY init-db.sh /docker-entrypoint-initdb.d
 
-HEALTHCHECK --interval=10s --timeout=10s --retries=10 CMD psql -c 'select 1' -d draftstore -U draftstore
+HEALTHCHECK --interval=10s --timeout=10s --retries=10 CMD psql --username postgres -c 'select 1' -d draftstore -U draftstore
```
rkondratowicz: isn't `-U draftstore` username as well?

docker-compose.yml
lgonczar: To be removed in all occurrences as a separate change

PR #49, opened 2017-10-10
Refactor - add service layer

rkondratowicz:
### Change description ###

Mostly in preparation for encryption (which will be added to `DraftService`)

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/functionalTest/java/uk/gov/hmcts/reform/draftstore/endpoint/v3/DeleteTest.java
+++ b/src/functionalTest/java/uk/gov/hmcts/reform/draftstore/endpoint/v3/DeleteTest.java
@@ -29,7 +29,7 @@ public class DeleteTest {
     @Autowired private MockMvc mockMvc;
 
     @MockBean private DraftService draftService;
-    @MockBean private AuthService authService; //NOPMD
+    @MockBean private AuthService authService; //NOPMD - mock declaration required
 
     @Test
     public void should_return_403_when_auth_exception_is_thrown() throws Exception {
--- a/src/functionalTest/java/uk/gov/hmcts/reform/draftstore/endpoint/v3/DeleteTest.java
+++ b/src/functionalTest/java/uk/gov/hmcts/reform/draftstore/endpoint/v3/DeleteTest.java
@@ -1,9 +1,7 @@
 package uk.gov.hmcts.reform.draftstore.endpoint.v3;
 
-import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
-import org.mockito.BDDMockito;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest;
 import org.springframework.boot.test.mock.mockito.MockBean;
@@ -11,17 +9,14 @@ import org.springframework.http.MediaType;
 import org.springframework.test.context.junit4.SpringRunner;
 import org.springframework.test.web.servlet.MockMvc;
 import org.springframework.test.web.servlet.ResultActions;
-import uk.gov.hmcts.reform.draftstore.data.DraftStoreDAO;
-import uk.gov.hmcts.reform.draftstore.domain.Draft;
-import uk.gov.hmcts.reform.draftstore.endpoint.v3.helpers.SampleData;
+import uk.gov.hmcts.reform.draftstore.exception.AuthorizationException;
 import uk.gov.hmcts.reform.draftstore.service.AuthService;
+import uk.gov.hmcts.reform.draftstore.service.DraftService;
 import uk.gov.hmcts.reform.draftstore.service.UserAndService;
 
-import java.util.Optional;
-
-import static org.mockito.Matchers.anyInt;
+import static org.mockito.BDDMockito.willThrow;
+import static org.mockito.Matchers.any;
 import static org.mockito.Matchers.anyString;
-import static org.mockito.Matchers.eq;
 import static org.springframework.http.HttpHeaders.AUTHORIZATION;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.delete;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/hmcts/reform/draftstore/endpoint/v3/DraftController.java
```diff
@@ -123,17 +107,9 @@ public DraftList readAll(
         @RequestHeader(SERVICE_HEADER) String serviceHeader,
         @RequestBody @Valid UpdateDraft updatedDraft
     ) {
-        UserAndService userAndService = authService.authenticate(authHeader, serviceHeader);
```
lgonczar: To me this line makes it a little bit clearer, as it's not obvious what authService.authenticate returns. Just to be considered.

src/functionalTest/java/uk/gov/hmcts/reform/draftstore/endpoint/v3/GetAllTest.java
```diff
@@ -29,18 +30,14 @@
     @Autowired
     private MockMvc mockMvc;
 
-    @MockBean private DraftStoreDAO draftRepo;
-    @MockBean private AuthService authService;
+    @MockBean private DraftService draftService;
+    @MockBean private AuthService authService; //NOPMD
```
timja-kainos: no pmd?

src/main/java/uk/gov/hmcts/reform/draftstore/service/DraftService.java
```diff
@@ -0,0 +1,88 @@
...
+
+import java.util.List;
+import java.util.Objects;
+
+@Service
+public class DraftService {
+
+    private final DraftStoreDAO draftRepo;
+
+    public DraftService(DraftStoreDAO draftRepo) {
```
lgonczar: Will the repo be injected here?

src/main/java/uk/gov/hmcts/reform/draftstore/service/DraftService.java
```diff
@@ -0,0 +1,88 @@
...
+
+    public Draft read(String id, UserAndService userAndService) {
+        return draftRepo
+            .read(toInternalId(id))
+            .filter(draft -> Objects.equals(draft.userId, userAndService.userId))
+            .filter(draft -> Objects.equals(draft.service, userAndService.service))
+            .orElseThrow(() -> new NoDraftFoundException());
+    }
+
+    public DraftList read(UserAndService userAndService, Integer after, int limit) {
```
lgonczar: Is there a reason why `after` is of type Integer?

src/functionalTest/java/uk/gov/hmcts/reform/draftstore/endpoint/v3/DeleteTest.java
```diff
@@ -33,71 +28,30 @@
...
-            .given(authService.authenticate(existingDraft.userId, existingDraft.service))
-            .willReturn(
-                new UserAndService(
-                    existingDraft.userId,
-                    existingDraft.service
-                )
-            );
-    }
+    @MockBean private DraftService draftService;
+    @MockBean private AuthService authService; //NOPMD
```
rkondratowicz: pmd complains about unused field, but it has to be there, we need a mock.

src/main/java/uk/gov/hmcts/reform/draftstore/service/DraftService.java
```diff
@@ -0,0 +1,88 @@
...
+    public void delete(String id, UserAndService userAndService) {
+        draftRepo
+            .read(toInternalId(id))
+            .ifPresent(d -> {
+                assertCanEdit(d, userAndService);
+                draftRepo.delete(toInternalId(id));
+            });
+    }
+
+    private void assertCanEdit(Draft draft, UserAndService userAndService) {
```
rlewan: Logic of checking if user and service match the record is again duplicated here. I would suggest to do all of these checks on query level, or have a new component which performs the check.

src/main/java/uk/gov/hmcts/reform/draftstore/service/DraftService.java
rkondratowicz: paging is optional. the type hasn't changed in this refactoring.

src/main/java/uk/gov/hmcts/reform/draftstore/service/DraftService.java
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * Converts external API id to internally used id.
+     */
+    private int toInternalId(String apiId) {
+        try {
+            // currently database ID is an int
+            return Integer.parseInt(apiId);
+        } catch (NumberFormatException exc) {
+            throw new NoDraftFoundException();
```
rlewan: `404` doesn't seem suitable for an unparsable parameter. I would suggest a `400` or `500` instead.

src/functionalTest/java/uk/gov/hmcts/reform/draftstore/endpoint/v3/GetAllTest.java
rkondratowicz: pmd complains about unused field, but it has to be there, we need a mock.

src/test/java/uk/gov/hmcts/reform/draftstore/service/draftservice/BaseTest.java
```diff
@@ -0,0 +1,43 @@
...
+import uk.gov.hmcts.reform.draftstore.domain.Draft;
+import uk.gov.hmcts.reform.draftstore.service.DraftService;
+import uk.gov.hmcts.reform.draftstore.service.UserAndService;
+
+import java.time.ZonedDateTime;
+import java.util.Optional;
+
+import static org.mockito.Matchers.anyInt;
+
+public class BaseTest {
```
lgonczar: We could rename it to indicate that it's only base test for a certain type of test classes (it contains draft service and draft repo).

How about `BaseControllerTest`?

src/functionalTest/java/uk/gov/hmcts/reform/draftstore/endpoint/v3/GetAllTest.java
lgonczar: would it still work if NOPMD was followed by explanation?

src/test/java/uk/gov/hmcts/reform/draftstore/service/draftservice/BaseTest.java
rkondratowicz: notice it's in its own package

src/main/java/uk/gov/hmcts/reform/draftstore/service/DraftService.java
lgonczar: I think 404 is the right code. The exception may be misleading, but what's really happening here is we're getting an ID of a non-existing draft.

src/test/java/uk/gov/hmcts/reform/draftstore/service/draftservice/BaseTest.java
```diff
@@ -0,0 +1,43 @@
...
+            "abc",
+            "serviceA",
+            "{}",
+            "some_type",
+            ZonedDateTime.now(),
+            ZonedDateTime.now()
+        );
+    }
+
+    void thereExists(Draft draft) {
```
lgonczar: This reads quite nicely in the tests 👍

src/test/java/uk/gov/hmcts/reform/draftstore/service/draftservice/DeleteTest.java
```diff
@@ -0,0 +1,47 @@
...
+import uk.gov.hmcts.reform.draftstore.service.UserAndService;
+
+import static org.assertj.core.api.Assertions.catchThrowable;
+import static org.assertj.core.api.Java6Assertions.assertThat;
+
+@RunWith(MockitoJUnitRunner.class)
+public class DeleteTest extends BaseTest {
+
+    @Test
+    public void should_throw_an_exception_when_trying_to_remove_somebody_elses_draft() throws Exception {
```
lgonczar: We could add a case for trying to remove the same person's draft for a different service and a case when we successfully delete an existing draft (we could verify repository call here).

src/functionalTest/java/uk/gov/hmcts/reform/draftstore/endpoint/v3/GetAllTest.java
rkondratowicz: done

src/test/java/uk/gov/hmcts/reform/draftstore/service/draftservice/ReadManyTest.java
```diff
@@ -0,0 +1,48 @@
...
+import uk.gov.hmcts.reform.draftstore.service.UserAndService;
+
+import static java.util.Collections.emptyList;
+import static java.util.Collections.singletonList;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Matchers.anyInt;
+import static org.mockito.Matchers.anyString;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ReadManyTest extends BaseTest {
```
lgonczar: It may be worth checking if the service called the repo with the right arguments.

src/test/java/uk/gov/hmcts/reform/draftstore/service/draftservice/ReadSingleTest.java
```diff
@@ -0,0 +1,61 @@
...
+import org.junit.runner.RunWith;
+import org.mockito.runners.MockitoJUnitRunner;
+import uk.gov.hmcts.reform.draftstore.exception.NoDraftFoundException;
+import uk.gov.hmcts.reform.draftstore.service.UserAndService;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.catchThrowable;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ReadSingleTest extends BaseTest {
```
lgonczar: We could have one more test checking what we return when the draft exists and is requested for the right (user, service) pair.

src/test/java/uk/gov/hmcts/reform/draftstore/service/draftservice/BaseTest.java
lgonczar: right 👍

src/main/java/uk/gov/hmcts/reform/draftstore/service/DraftService.java
rkondratowicz: I'll move those two functions (one predicate, one check throwing an exception) to a util class in a future pull request.

rlewan: 👍

src/main/java/uk/gov/hmcts/reform/draftstore/service/DraftService.java
rlewan: For me that's hiding the problem, making troubleshooting harder. Validation shouldn't let unparsable parameters through and reaching this code section should mean that there is a bug in the application.

rkondratowicz: The background for this is that a requirement was added for the ID to be a string (on API level).  
@rlewan had a good point that we can keep it a string and add a regex (documented by Swagger).

Will investigate separately.

src/test/java/uk/gov/hmcts/reform/draftstore/service/draftservice/ReadSingleTest.java
rkondratowicz: added extra tests

src/main/java/uk/gov/hmcts/reform/draftstore/service/DraftService.java
lgonczar: I don't see how it makes it harder to troubleshoot. If the API provides a non-existing ID, it gets a 404 in every case. There's no reason to validate whether it's an integer or not. We don't require it to be an int and it's not an error in the service if someone provides something else.

PR #50, opened 2017-10-11
add crypto service

rkondratowicz:
### Change description ###
Just the service with tests. Not used by API yet.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/draftstore/service/crypto/CryptoService.java
+++ b/src/main/java/uk/gov/hmcts/reform/draftstore/service/crypto/CryptoService.java
@@ -70,8 +70,8 @@ public class CryptoService {
     }
 
     private static SecretKey deriveKey(String secret) throws NoSuchAlgorithmException {
-        byte[] hashedPassword = MessageDigest.getInstance("SHA-512").digest(secret.getBytes());
-        byte[] key = Arrays.copyOfRange(hashedPassword, 0, KEY_LENGTH.bytes());
+        byte[] hashedSecret = MessageDigest.getInstance("SHA-512").digest(secret.getBytes());
+        byte[] key = Arrays.copyOfRange(hashedSecret, 0, KEY_LENGTH.bytes());
 
         return new SecretKeySpec(key, "AES");
     }
--- a/src/main/java/uk/gov/hmcts/reform/draftstore/service/crypto/CryptoService.java
+++ b/src/main/java/uk/gov/hmcts/reform/draftstore/service/crypto/CryptoService.java
@@ -1,0 +1,85 @@
+package uk.gov.hmcts.reform.draftstore.service.crypto;
+
+import com.google.common.base.Strings;
+import com.google.common.primitives.Bytes;
+import uk.gov.hmcts.reform.draftstore.utils.IllegalArgument;
+
+import java.security.GeneralSecurityException;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+import java.util.Arrays;
+import javax.crypto.Cipher;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.GCMParameterSpec;
+import javax.crypto.spec.SecretKeySpec;
+
+import static com.google.common.base.Charsets.UTF_8;
+
+public class CryptoService {
+
+    private static final String CIPHER = "AES/GCM/NoPadding";
+    private static final Length KEY_LENGTH = Length.BITS_128;
+    private static final Length IV_LENGTH = Length.BITS_96;
+    private static final Length GMC_TAG_LENGTH = Length.BITS_128;
+
+    public static byte[] encrypt(String input, String password) {
+
+        IllegalArgument.throwIf(() -> input == null, "Input can't be null");
+        IllegalArgument.throwIf(() -> Strings.isNullOrEmpty(password), "Password can't be empty");
+
+        try {
+            Cipher cipher = Cipher.getInstance(CIPHER);
+            cipher.init(
+                Cipher.ENCRYPT_MODE,
+                deriveKey(password),
+                new GCMParameterSpec(GMC_TAG_LENGTH.bits(), generateIv())
+            );
+
+            return Bytes.concat(
+                cipher.getIV(),
+                cipher.doFinal(input.getBytes(UTF_8))
+            );
+
+        } catch (GeneralSecurityException exc) {
+            throw new GeneralSecurityRuntimeException(exc);
+        }
+    }
+
+    public static String decrypt(byte[] input, String password) {
+
+        IllegalArgument.throwIf(() -> input == null, "Input can't be null");
+        IllegalArgument.throwIf(() -> Strings.isNullOrEmpty(password), "Password can't be empty");
+
+        try {
+            byte[] iv = Arrays.copyOfRange(input, 0, IV_LENGTH.bytes());
+            byte[] cipherText = Arrays.copyOfRange(input, IV_LENGTH.bytes(), input.length);
+
+            Cipher cipher = Cipher.getInstance(CIPHER);
+            cipher.init(
+                Cipher.DECRYPT_MODE,
+                deriveKey(password),
+                new GCMParameterSpec(GMC_TAG_LENGTH.bits(), iv)
+            );
+
+            return new String(cipher.doFinal(cipherText), UTF_8);
+
+        } catch (GeneralSecurityException exc) {
+            throw new GeneralSecurityRuntimeException(exc);
+        }
+    }
+
+    private static SecretKey deriveKey(String password) throws NoSuchAlgorithmException {
+        byte[] hashedPassword = MessageDigest.getInstance("SHA-512").digest(password.getBytes());
+        byte[] key = Arrays.copyOfRange(hashedPassword, 0, KEY_LENGTH.bytes());
+
+        return new SecretKeySpec(key, "AES");
+    }
+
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/hmcts/reform/draftstore/service/crypto/CryptoService.java
```diff
@@ -0,0 +1,85 @@
...
+
+    private static final String CIPHER = "AES/GCM/NoPadding";
+    private static final Length KEY_LENGTH = Length.BITS_128;
+    private static final Length IV_LENGTH = Length.BITS_96;
+    private static final Length GMC_TAG_LENGTH = Length.BITS_128;
+
+    public static byte[] encrypt(String input, String password) {
+
+        IllegalArgument.throwIf(() -> input == null, "Input can't be null");
+        IllegalArgument.throwIf(() -> Strings.isNullOrEmpty(password), "Password can't be empty");
```
timja-kainos: Any requirements on the password?

src/main/java/uk/gov/hmcts/reform/draftstore/service/crypto/CryptoService.java
```diff
@@ -0,0 +1,85 @@
...
+import static com.google.common.base.Charsets.UTF_8;
+
+public class CryptoService {
+
+    private static final String CIPHER = "AES/GCM/NoPadding";
+    private static final Length KEY_LENGTH = Length.BITS_128;
+    private static final Length IV_LENGTH = Length.BITS_96;
+    private static final Length GMC_TAG_LENGTH = Length.BITS_128;
+
+    public static byte[] encrypt(String input, String password) {
```
timja-kainos: Is there a library that could this for us and be less risky for us to implement?

rkondratowicz: There's not much going on here, we just provide configuration for `javax.Crypto` (key length, IV length, AES mode etc).

The only custom things is that IV is concatenated with cipher text.

src/main/java/uk/gov/hmcts/reform/draftstore/service/crypto/CryptoService.java
lgonczar: It would be a good idea add some validation of the password (length) to request validation. Not critical, though, as our clients are other HMCTS services (responsible for using strong passwords), not individuals.

timja-kainos: probably a min length would be enough...

rkondratowicz: I'd rather validate that on API level

src/main/java/uk/gov/hmcts/reform/draftstore/service/crypto/CryptoService.java
```diff
@@ -0,0 +1,85 @@
...
+                new GCMParameterSpec(GMC_TAG_LENGTH.bits(), generateIv())
+            );
+
+            return Bytes.concat(
+                cipher.getIV(),
+                cipher.doFinal(input.getBytes(UTF_8))
+            );
+
+        } catch (GeneralSecurityException exc) {
+            throw new GeneralSecurityRuntimeException(exc);
```
lgonczar: Worth adding a message, saying we failed to encrypt the data. I imagine that the message in exc may be quite enigmatic.

src/main/java/uk/gov/hmcts/reform/draftstore/service/crypto/CryptoService.java
rkondratowicz: Alternatively we can require a 128bit key to be sent from the client. Instead of deriving it from the password

src/main/java/uk/gov/hmcts/reform/draftstore/service/crypto/CryptoService.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+    }
+
+    public static String decrypt(byte[] input, String password) {
+
+        IllegalArgument.throwIf(() -> input == null || input.length == 0, "Input can't be empty");
```
lgonczar: You could consider using Strings.isNullOrEmpty, like you did below. Looks a bit nicer.

rkondratowicz: it's a byte array here

src/main/java/uk/gov/hmcts/reform/draftstore/service/crypto/CryptoService.java
```diff
@@ -0,0 +1,85 @@
...
+            byte[] cipherText = Arrays.copyOfRange(input, IV_LENGTH.bytes(), input.length);
+
+            Cipher cipher = Cipher.getInstance(CIPHER);
+            cipher.init(
+                Cipher.DECRYPT_MODE,
+                deriveKey(password),
+                new GCMParameterSpec(GMC_TAG_LENGTH.bits(), iv)
+            );
+
+            return new String(cipher.doFinal(cipherText), UTF_8);
```
lgonczar: Maybe we should rely on the encoding set up for the application? It's probably going to still be UTF_8, but at least there would be one place where we specify it. What do you think?

src/test/java/uk/gov/hmcts/reform/draftstore/service/crypto/CryptoServiceTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.junit.Test;
+
+import java.util.stream.Stream;
+import javax.crypto.AEADBadTagException;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.catchThrowable;
+
+public class CryptoServiceTest {
```
KarolJastrzebski: Missing assertions for:
- plain == decrypted
- plain != encrypted

src/main/java/uk/gov/hmcts/reform/draftstore/service/crypto/CryptoService.java
```diff
@@ -0,0 +1,85 @@
...
+
+        } catch (GeneralSecurityException exc) {
+            throw new GeneralSecurityRuntimeException(exc);
+        }
+    }
+
+    private static SecretKey deriveKey(String password) throws NoSuchAlgorithmException {
+        byte[] hashedPassword = MessageDigest.getInstance("SHA-512").digest(password.getBytes());
```
lgonczar: I'm wondering if adding some salt here would increase security. If the user has multiple drafts and some drafts are the same, they would also look the same encrypted. On the other hand, does it matter?

rkondratowicz: The cipher text will be different. Random IV is used every time.

src/main/java/uk/gov/hmcts/reform/draftstore/service/crypto/Length.java
```diff
@@ -0,0 +1,21 @@
+package uk.gov.hmcts.reform.draftstore.service.crypto;
+
+public enum Length {
```
lgonczar: good idea 👍

src/main/java/uk/gov/hmcts/reform/draftstore/service/crypto/CryptoService.java
lgonczar: Those passwords could be named `secrets`. To us they're secret pieces of information, based on which we will encrypt the data.

src/main/java/uk/gov/hmcts/reform/draftstore/service/crypto/CryptoService.java
lgonczar: Right, you add it to the message.

src/test/java/uk/gov/hmcts/reform/draftstore/service/crypto/CryptoServiceTest.java
rkondratowicz: added

src/main/java/uk/gov/hmcts/reform/draftstore/service/crypto/CryptoService.java
lgonczar: 👍

src/test/java/uk/gov/hmcts/reform/draftstore/service/crypto/CryptoServiceTest.java
lgonczar: We could also add a test for checking if an exception would be thrown if we used an empty password or an empty input (although maybe we should allow for input to be empty) and a test checking if encrypted input of a very short length is long enough in the encrypted form.

src/main/java/uk/gov/hmcts/reform/draftstore/service/crypto/CryptoService.java
```diff
@@ -0,0 +1,85 @@
...
+
+            return new String(cipher.doFinal(cipherText), UTF_8);
+
+        } catch (GeneralSecurityException exc) {
+            throw new GeneralSecurityRuntimeException(exc);
+        }
+    }
+
+    private static SecretKey deriveKey(String secret) throws NoSuchAlgorithmException {
+        byte[] hashedPassword = MessageDigest.getInstance("SHA-512").digest(secret.getBytes());
```
timja-kainos: hashedSecret?

PR #51, opened 2017-10-12
Optional draft encryption

rkondratowicz:
### Change description ###

Makes it possible to draft-store clients to encrypt drafts.

As it is not mandatory during the transitions stage, database now has two columns
to store draft (in either plaintext or encrypted form) but only one column is
being populated:

For write operations, if header containg secret is set, document is encrypted
before storing in db using key derived from provided secret.

For read operations, if db row contains encrypted document, secret sent by the clients
will be used to decrypt the document.

Added a new set of models to be used by repository layer. Those models contain
data for both columns, repo just stores/retrieves the data.
The data transformation is handled by the service layer (added recently).

### DIFF
--- a/src/functionalTest/java/uk/gov/hmcts/reform/draftstore/data/DraftStoreDAOTest.java
+++ b/src/functionalTest/java/uk/gov/hmcts/reform/draftstore/data/DraftStoreDAOTest.java
@@ -9,14 +9,13 @@ import org.springframework.dao.DataIntegrityViolationException;
 import org.springframework.test.context.ActiveProfiles;
 import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
 import org.springframework.transaction.annotation.Transactional;
-import uk.gov.hmcts.reform.draftstore.domain.Draft;
+import uk.gov.hmcts.reform.draftstore.data.model.Draft;
 import uk.gov.hmcts.reform.draftstore.domain.SaveStatus;
 import uk.gov.hmcts.reform.draftstore.exception.NoDraftFoundException;
 
 import java.sql.SQLException;
 import java.util.List;
 
-import static java.util.Collections.singletonMap;
 import static org.assertj.core.api.Assertions.assertThat;
 import static uk.gov.hmcts.reform.draftstore.domain.SaveStatus.Created;
 import static uk.gov.hmcts.reform.draftstore.domain.SaveStatus.Updated;
--- a/src/main/java/uk/gov/hmcts/reform/draftstore/data/model/CreateDraft.java
+++ b/src/main/java/uk/gov/hmcts/reform/draftstore/data/model/CreateDraft.java
@@ -1,0 +1,18 @@
+package uk.gov.hmcts.reform.draftstore.data.model;
+
+public class CreateDraft {
+
+    public final String document;
+    public final byte[] encryptedDocument;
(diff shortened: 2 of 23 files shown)


src/main/java/uk/gov/hmcts/reform/draftstore/data/model/CreateDraft.java
```diff
@@ -0,0 +1,18 @@
...
+    public final Integer maxStaleDays;
+
+    // region constructor
+    public CreateDraft(String document, byte[] encryptedDocument, String type, Integer maxStaleDays) {
+        this.document = document;
+        this.encryptedDocument = encryptedDocument;
+        this.type = type;
+        this.maxStaleDays = maxStaleDays;
+    }
+    // endregion
```
yarpo: do you want to merge region's comments?

src/main/java/uk/gov/hmcts/reform/draftstore/endpoint/EndpointExceptionHandler.java
```diff
@@ -122,6 +123,15 @@
         );
     }
 
+    @ExceptionHandler(InvalidKeyException.class)
+    public ResponseEntity<ErrorResult> invalidCryptoKey(HttpServletRequest req, Exception exception) {
+        log.warn(exception.getMessage(), exception);
+        return new ResponseEntity<>(
+            new ErrorResult(BAD_ARGUMENT, singletonList(exception.getMessage())),
```
yarpo: I'd reformat this to:

```
new Err(arg1, arg2, arg3)
or
new Err(
   arg1, arg2, arg3
)
or
new Err(
   arg1, 
   arg2, 
   arg3
)
```

src/functionalTest/java/uk/gov/hmcts/reform/draftstore/data/DraftStoreDAOTest.java
```diff
@@ -9,14 +9,13 @@
 import org.springframework.test.context.ActiveProfiles;
 import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
 import org.springframework.transaction.annotation.Transactional;
-import uk.gov.hmcts.reform.draftstore.domain.Draft;
```
timja-kainos: General question on this PR:
What happens if the client sends a secret and the document isn't encrypted?
During the transition stage you wouldn't know if its encrypted or not

rkondratowicz: we don't make any changes to it and return plain text. On next update it will be encrypted though and plaintext db column will be set to null.

See `FromDbModelMapper` for the read part

rkondratowicz:
```
// only during transition stage
String documentToReturn =
    dbDraft.encryptedDocument == null
        ? dbDraft.document
        : CryptoService.decrypt(dbDraft.encryptedDocument, secret);
```

rkondratowicz: also, notice that this happens 'per draft', so it is possible to read a mix of old (plaintext) and new (encrypted) drafts from db for given user.

PR #53, opened 2017-10-16
Add support for primary and secondary secret

rkondratowicz:
### Change description ###

Support multiple secrets so that draft store clients can rotate their keys.
The same single header as before is used, secrets are separated with `,`.
Reads try primary then (if present) secondary secret. Writes always use primary key.


src/functionalTest/java/uk/gov/hmcts/reform/draftstore/endpoint/v3/GetAllTest.java
```diff
@@ -31,11 +32,15 @@
     private MockMvc mockMvc;
 
     @MockBean private DraftService draftService;
-    @MockBean private AuthService authService; //NOPMD - mock declaration required
```
lgonczar: Does PMD no longer complain about it?

rkondratowicz: AuthService is not used below, so the comment is not longer needed

src/main/java/uk/gov/hmcts/reform/draftstore/endpoint/EndpointExceptionHandler.java
```diff
@@ -132,6 +133,15 @@
         );
     }
 
+    @ExceptionHandler(SecretsException.class)
+    public ResponseEntity<ErrorResult> secrets(HttpServletRequest req, Exception exception) {
```
lgonczar: How about `invalidSecrets`?

src/main/java/uk/gov/hmcts/reform/draftstore/endpoint/v3/DraftController.java
```diff
@@ -98,8 +104,10 @@ public DraftList readAll(
         @RequestHeader(name = SECRET_HEADER, required = false) @Valid @Length(min = MIN_SECRET_LENGTH) String secretHeader,
         @RequestBody @Valid CreateDraft newDraft
     ) {
-        UserAndService userAndService = authService.authenticate(authHeader, serviceHeader, secretHeader);
-        int id = draftService.create(newDraft, userAndService);
+        Secrets secrets = Secrets.fromHeader(secretHeader);
```
lgonczar: We're repeating this sequence a few times:

- get secrets from header
- authenticate using auth header
- add secrets to UserAndService
- perform an action using UserAndService

I'm wondering if we could extract some common code. What do you think?

src/main/java/uk/gov/hmcts/reform/draftstore/service/UserAndService.java
```diff
@@ -1,18 +1,24 @@
...
-        this.secret = secret;
+        this.secrets = secrets;
     }
 
     public UserAndService(String userId, String service) {
-        this(userId, service, null);
+        this(userId, service, new Secrets(null, null));
+    }
+
+    public UserAndService withSecrets(Secrets secrets) {
```
lgonczar: 👍

src/main/java/uk/gov/hmcts/reform/draftstore/endpoint/v3/DraftController.java
rkondratowicz: Once we start using Spring Security this will be gone :)

src/main/java/uk/gov/hmcts/reform/draftstore/utils/Retry.java
```diff
@@ -0,0 +1,33 @@
...
+     *
+     * @return The first successful result of calling a function.
+     */
+    public static <T, R> R with(
+        List<T> params,
+        Function<T, R> function
+    ) {
+        IllegalArgument.throwIf(() -> params.isEmpty(), "params list must not be empty");
+
+        for (int i = 0; i < params.size() - 1; i++) {
```
lgonczar: As `params` is a list, we could just iterate over its elements.

src/main/java/uk/gov/hmcts/reform/draftstore/utils/Retry.java
```diff
@@ -0,0 +1,33 @@
...
+            try {
+                return function.apply(params.get(i));
+            } catch (RuntimeException exc) {
+                LOGGER.warn("Failed retry", exc);
```
lgonczar: This doesn't tell us much. As it's a general-purpose util, I wonder if there would be a way of providing more specific messages, like "Failed to decrypt the document using secret #1".

src/main/java/uk/gov/hmcts/reform/draftstore/utils/Retry.java
lgonczar: Ok, I see why.

rkondratowicz: The operations below are not performed on the last item though. Any suggestions?

src/main/java/uk/gov/hmcts/reform/draftstore/service/secrets/Secrets.java
```diff
@@ -0,0 +1,50 @@
...
+
+public class Secrets {
+
+    public static final int MIN_SECRET_LENGTH = 16;
+    public static final String SEPARATOR = ",";
+
+    public final String primary;
+    public final String secondary;
+
+    public Secrets(String primary, String secondary) {
```
lgonczar: Maybe secondary should be wrapped in Optional, so that it would be obvious this one is not required.

lgonczar: I take that back, seeing that both are optional.

src/main/java/uk/gov/hmcts/reform/draftstore/service/secrets/Secrets.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public static Secrets fromHeader(String header) {
+        if (Strings.isNullOrEmpty(header)) {
+            return new Secrets(null, null);
+        } else {
+            List<String> secrets = asList(header.split(SEPARATOR));
+            if (secrets.size() > 2) {
+                throw new SecretsException("Too many secrets. Max number is 2");
+            } else {
+                if (!secrets.stream().allMatch(s -> s.length() >= MIN_SECRET_LENGTH)) {
```
lgonczar: This if-else doesn't need to be nested. Can be just another `else`.

src/main/java/uk/gov/hmcts/reform/draftstore/service/secrets/Secrets.java
```diff
@@ -0,0 +1,50 @@
...
+                    throw new SecretsException("Min length for secret is " + MIN_SECRET_LENGTH);
+                } else {
+                    return new Secrets(
+                        secrets.get(0).trim(),
+                        Lists.safeGet(secrets, 1).map(s -> s.trim()).orElse(null)
```
lgonczar: No need for this check - we know there's an element under index 1.

As for trimming, it would be worth applying it before verifying the length, as well.

src/test/java/uk/gov/hmcts/reform/draftstore/service/mappers/ToDbModelMapperTest.java
```diff
@@ -12,20 +14,33 @@
...
     public void should_set_only_unencrypted_document_when_secret_is_NOT_passed() throws Exception {
-        result = ToDbModelMapper.toDb(createDraft(), null);
+        result = ToDbModelMapper.toDb(createDraft(), new Secrets(null, null));
 
         assertThat(result.document).isNotNull();
         assertThat(result.encryptedDocument).isNull();
     }
 
+    @Test
+    public void should_use_primary_secret_for_encryption() throws Exception {
```
lgonczar: Now that we may have two keys, we should check if:

 - the mapper will use the secondary key (with success) when the primary one fails
 - if the mapper will fail in the expected way when the secondary key fails

src/test/java/uk/gov/hmcts/reform/draftstore/utils/RetryTest.java
```diff
@@ -0,0 +1,46 @@
...
+    public void should_return_result_of_calling_an_action_with_second_param_if_first_one_failed() throws Exception {
+        int result = Retry.with(asList(0, 24), n -> failOnZero(n));
+        assertThat(result).isEqualTo(24);
+    }
+
+    @Test
+    public void should_throw_an_exception_when_action_fails_for_all_params() throws Exception {
+        assertThatThrownBy(
+            () -> Retry.with(asList(0, 0, 0), n -> failOnZero(n))
+        ).isInstanceOf(RuntimeException.class);
```
lgonczar: It would be ideal to check if it's the exception thrown by the function wrapped by this Retry. Or maybe the tool should throw a higher-level exception, with this exception as the cause?

src/functionalTest/java/uk/gov/hmcts/reform/draftstore/endpoint/v3/GetAllTest.java
lgonczar: 👍

src/main/java/uk/gov/hmcts/reform/draftstore/service/secrets/Secrets.java
rkondratowicz: We do need it, the can be just one, primary secret in the header.  
There's a test for that.

Good point about trimming 👍

lgonczar: Right, my bad

src/main/java/uk/gov/hmcts/reform/draftstore/service/secrets/Secrets.java
```diff
@@ -0,0 +1,46 @@
...
+                throw new SecretsException("Min length for secret is " + MIN_SECRET_LENGTH);
+            } else {
+                return new Secrets(
+                    secrets.get(0),
+                    Lists.safeGet(secrets, 1).orElse(null)
+                );
+            }
+        }
+    }
+}
```
timja-kainos: perhaps add a toString method that doesn't allow logging of these values (I know there's not one here now but someone may add one later and not think about it)

something like:
```
toString() { return "These values are secret"}
```

src/main/java/uk/gov/hmcts/reform/draftstore/service/secrets/Secrets.java
```diff
@@ -0,0 +1,46 @@
...
+
+    public final String primary;
+    public final String secondary;
+
+    public Secrets(String primary, String secondary) {
+        this.primary = primary;
+        this.secondary = secondary;
+    }
+
+    public static Secrets fromHeader(String header) {
```
KarolJastrzebski: :+1:

src/main/java/uk/gov/hmcts/reform/draftstore/service/secrets/Secrets.java
lgonczar: Or `toString() { return Secrets(numer of secrets); }`

src/main/java/uk/gov/hmcts/reform/draftstore/service/mappers/FromDbModelMapper.java
```diff
@@ -2,18 +2,30 @@
...
+        if (dbDraft.encryptedDocument == null) {
+            documentToReturn = dbDraft.document;
+        } else {
+            try {
+                documentToReturn = CryptoService.decrypt(dbDraft.encryptedDocument, secrets.primary);
+            } catch (Exception exc) {
+                if (secrets.secondary != null) {
+                    documentToReturn = CryptoService.decrypt(dbDraft.encryptedDocument, secrets.secondary);
+                } else {
+                    throw exc;
```
lgonczar: Just to consider - the exception thrown here may be hard to understand, so we could wrap it in our exception, saying "Failed to decrypt the document from the draft".

lgonczar: Oh, that could complicate converting exceptions to responses. Maybe not worth it, then.

src/test/java/uk/gov/hmcts/reform/draftstore/service/mappers/ToDbModelMapperTest.java
rkondratowicz: done.

PR #61, opened 2017-10-24
Make gatling config file more consistent

rkondratowicz:

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/gatling/scala/uk/gov/hmcts/reform/draftstore/actions/setup/LeaseServiceToken.scala
+++ b/src/gatling/scala/uk/gov/hmcts/reform/draftstore/actions/setup/LeaseServiceToken.scala
@@ -11,7 +11,7 @@ import scala.util.Random
 
 object LeaseServiceToken {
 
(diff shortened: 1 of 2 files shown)


src/gatling/scala/uk/gov/hmcts/reform/draftstore/actions/setup/LeaseServiceToken.scala
```diff
@@ -23,7 +23,7 @@ object LeaseServiceToken {
     feed(serviceNameFeeder)
       .exec(
         http("Lease service token")
-          .post(url)
+          .post(s2sUrl + "/testing-support/lease")
```
rlewan: What do you think about [interpolating](https://docs.scala-lang.org/overviews/core/string-interpolation.html) that string instead?

rkondratowicz: I use it in other places, prefer '+' here.

PR #66, opened 2017-10-28
Use TestContainers to run functional tests

geek-soft:

None

### Change description ###

Change introduces TestContainers to start database instance in docker environment prior to run functional tests. That should simplify development process by not requiring database to be 'up and running'  when functional tests are run.

TestContainers doc: https://www.testcontainers.org/usage/database_containers.html#benefits

```
[ ] Yes
[x] No
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 sudo: required
 language: java
-services:
-  - docker
 jdk:
   - oraclejdk8
 os:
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -1,7 +1,5 @@
 sudo: required
```
timja-kainos: @geek-soft this also isn't required anymore, it means a full vm is used rather than a docker container.

PR #70, opened 2017-11-07
Fix publish swagger ci build step

rkondratowicz:
Bring back docker to travis (removed in https://github.com/hmcts/draft-store/pull/66) and increase sleep time

### DIFF
--- a/publish-swagger-docs.sh
+++ b/publish-swagger-docs.sh
@@ -9,7 +9,7 @@ then
 fi
 
 docker-compose up -d
-sleep 10
+sleep 15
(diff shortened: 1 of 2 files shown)


publish-swagger-docs.sh
```diff
@@ -9,7 +9,7 @@ then
 fi
 
 docker-compose up -d
-sleep 10
+sleep 15
```
lgonczar: Did the build fail because the sleep was too short?

rkondratowicz: it does for me locally

PR #73, opened 2017-11-09
Add route for deleting all user's drafts

rkondratowicz:

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/draftstore/data/DraftStoreDAO.java
+++ b/src/main/java/uk/gov/hmcts/reform/draftstore/data/DraftStoreDAO.java
@@ -180,6 +180,15 @@ public class DraftStoreDAO {
         );
     }
 
+    public void deleteAll(String userId, String service) {
+        jdbcTemplate.update(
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/reform/draftstore/service/DraftService.java
```diff
@@ -73,6 +73,10 @@ public void delete(String id, UserAndService userAndService) {
             });
     }
 
+    public void deleteAll(UserAndService userAndService) {
```
lgonczar: this method should be tested

src/main/java/uk/gov/hmcts/reform/draftstore/data/DraftStoreDAO.java
```diff
@@ -180,6 +180,15 @@ public void delete(int id) {
         );
     }
 
+    public void deleteAll(String userId, String service) {
```
lgonczar: we could add a functional test for this

rkondratowicz: done

PR #82, opened 2017-11-14
Fix getting swagger doc on travis

rkondratowicz:
Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```


publish-swagger-docs.sh
```diff
@@ -8,6 +8,7 @@ then
   exit 0
 fi
 
+./gradlew clean installDist
```
timja-kainos: shouldn't this be done in the travis script section?
https://github.com/hmcts/draft-store/blob/master/.travis.yml#L24

rkondratowicz: imho, it feel more 'self contained' when it's here

doncem:
I vote for moving to `.travis.yml` as well as precondition. Final result will be just a step of after success:
```yml
after_success:
  - codacy publish
  - test $TRAVIS_BRANCH = "master" && ./gradlew clean installDist && ./publish-swagger-docs.sh
```

rkondratowicz:
how about we merge it as is and see if it fixes the issue and then refactor to using 'after-success' in travis file?

PR #86, opened 2017-11-22
Handle short fake auth headers in idam client stub

rkondratowicz:
**Does this PR introduce a breaking change?**
```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/draftstore/service/idam/IdamClientStub.java
+++ b/src/main/java/uk/gov/hmcts/reform/draftstore/service/idam/IdamClientStub.java
@@ -2,10 +2,13 @@ package uk.gov.hmcts.reform.draftstore.service.idam;
 
 public class IdamClientStub implements IdamClient {
 
-    private static final int MAX_USER_ID_LENGTH = 256; // enforced by database column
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/draftstore/service/idam/IdamClientStub.java
```diff
@@ -2,10 +2,13 @@
...
-    private static final int MAX_USER_ID_LENGTH = 256; // enforced by database column
+    static final int MAX_USER_ID_LENGTH = 256; // enforced by database column
 
     @Override
     public User getUserDetails(String authHeader) {
-        return new User(authHeader.substring(0, MAX_USER_ID_LENGTH), "example@example.com");
+        String id = authHeader.substring(0, Math.min(authHeader.length(), MAX_USER_ID_LENGTH));
+        String email = "example@example.com";
+
+        return new User(id, email);
```
KarolJastrzebski: This length capping is used only in stub. Wouldn't be better to move it to User's constructor?

rkondratowicz: it's stub specific

PR #91, opened 2017-12-01
Extract swagger documentation to separate module

doncem:
### Change description ###

Extract swagger documentation to separate module. Affected changes:
- get rid of swagger incurred extra libraries from main application
- docker runs same as usual just building mirror of application including swagger configuration

Cons:
- Would be good to use separate port. Could not make it work
- Would be better to extract port to (environment) properties. Could not make it work for swagger module - still kept on using main application value instead of overridden one

```
[ ] Yes
[x] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,37 +1,44 @@
 plugins {
-  id 'application'
-  id 'checkstyle'
   id 'io.spring.dependency-management' version '1.0.3.RELEASE'
   id 'org.springframework.boot' version '1.5.7.RELEASE'
   id 'org.owasp.dependencycheck' version '1.4.5.1'
   id 'com.github.ben-manes.versions' version '0.15.0'
   id 'org.sonarqube' version '2.5'
-  id 'jacoco'
-  id 'pmd'
 }
 
-group 'uk.gov.hmcts.reform'
+allprojects {
+  apply plugin: 'application'
+  apply plugin: 'jacoco'
+  apply plugin: 'checkstyle'
+  apply plugin: 'pmd'
+
+  group 'uk.gov.hmcts.reform'
+
+  checkstyle.toolVersion = '7.2'
+  checkstyle.configFile = new File(rootDir, "checkstyle.xml")
+
+  // https://jeremylong.github.io/DependencyCheck/dependency-check-gradle/configuration.html
+  dependencyCheck {
+    // Specifies if the build should be failed if a CVSS score above a specified level is identified.
+    // range of 0-10 fails the build, anything greater and it doesn't fail the build
+    failBuildOnCVSS = System.getProperty('dependencyCheck.failBuild') == 'true' ? 0 : 11
+    suppressionFile = 'dependency-check-suppressions.xml'
+  }
+
+  sourceCompatibility = 1.8
+  targetCompatibility = 1.8
+
+  repositories {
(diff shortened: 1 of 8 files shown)


rkondratowicz:
This also stops the app from serving swagger spec when it's running.
Is that what we want @jasonpaige ?

jasonpaige:
I'm assuming in dev you'd probably still want to be able to run the swagger spec?

build.gradle
```diff
@@ -70,12 +76,13 @@ dependencies {
...
   compile group: 'org.flywaydb', name: 'flyway-core', version: '4.1.2'
 
-  compile group: 'io.springfox', name: 'springfox-swagger2', version: versions.springfoxSwagger
-  compile group: 'io.springfox', name: 'springfox-swagger-ui', version: versions.springfoxSwagger
+  compileOnly group: 'io.swagger', name: 'swagger-annotations', version: '1.5.17'
 
   compile group: 'com.fasterxml.jackson.core', name: 'jackson-databind', version: versions.jackson
   compile group: 'com.fasterxml.jackson.datatype', name: 'jackson-datatype-jsr310', version: versions.jackson
 
+  compile group: 'com.google.guava', name: 'guava', version: '20.0'
```
timja-kainos: what is guava needed for here?

doncem: for [CryptoService](https://github.com/hmcts/draft-store/blob/master/src/main/java/uk/gov/hmcts/reform/draftstore/service/crypto/CryptoService.java)

doncem:
Can boot up `docker-compose up draft-store-swagger` and will have all the previous endpoints present whereas `docker-compose up draft-store-api` - all but swagger

jasonpaige:
that approach works for me as long as it's documented. How about others?

jasonpaige:
As for the maintainer to cover codacy, just suppress the error on the project for now (need to login to codacy i think)

rkondratowicz:
is it worth it? do we want to change all the spring services now?
Are we not overcomplicating it?

build.gradle
rkondratowicz: and it previously wasn't? O_o

timja-kainos:
also not sure if its worth it

build.gradle
doncem: beats me. only can assume was injected together with springfox - did not check maven repo for dependencies, but guava pom has smart plugin injection and cannot **just** grab `google.common` packages ¯\\\_(ツ)_/¯

if it is true (about springfox) this is why it's not great to have such libraries in main source code just for swagger docs. docs in general could be separated

doncem:
Benefit: no extra libs in source code
"Worthness": depends on the view. can easily scratch it to 'none' - why spend time separating when it all works as is.

When it comes down to service / source code quality, I'd rather keep away such _extra_ code separate to minify the main project to bare bones ensuring everything is actually needed for production. As expected here - as soon as moved away swagger libraries - crypto service stopped working. Or might be some local issue as it was the first time I compiled draft store

We have centralised api spec place which is generated during CI process. But if we agree on leaving as it was - can just close the PR. No fuss

P.S. Jason, it's your app Codacy report ran against :D

doncem:
Closing according to failure of worthiness and lack of documented proof how/whether reform microservices should serve API documentation.

PR #93, opened 2017-12-13
Add GDPR note requested by Data and Security Architecture

jasonpaige:
### Change description ###
It's a legal requirement that we make it clear to operators of this
service that they must _not_ breach privacy and view the persisted data,
through any means.

```
[ ] Yes
[x] No
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,12 @@ Run the application by executing:
 ./gradlew bootRun
 ```
 
+### Operational note
+When operating this service with live data, no one should under any circumstances view the encrypted data. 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -27,6 +27,12 @@ Run the application by executing:
 ./gradlew bootRun
 ```
 
+### Operational note
```
timja-kainos: even operationally to support the user, at the request of the user?

jasonpaige: I think at that point it gets a little more complex in terms of business process as it would need to be qualified staff and some form of official record that the citizen had given express permission.

It's a valid requirement I think we should bring up with Data and Security to see how we can facilitate it.

timja-kainos: would be great if this was explored as do we have these problems...

jasonpaige: copied you into an email about it 👍

PR #96, opened 2017-12-14
Update Jenkinsfile

constantinpr:
### Change description ###

Code snippet added so that Jenkins build pushes the relevant info for grafana to influx db 

```
[ ] Yes
[x ] No
```


Jenkinsfile
```diff
@@ -65,6 +65,10 @@ node {
     archiveArtifacts 'build/reports/**/*.html'
     notifyBuildFailure channel: channel
     throw err
-  }
+  } finally {
+        step([$class: 'InfluxDbPublisher',
+               customProjectName: 'CMC Draftstore',
```
jasonpaige: I'm not sure where this value is used but draft-store is now a common component and it may be best to have a naming convention in influx which matches the repo name?

timja-kainos: yes, please remove the CMC from this

constantinpr: Removed the CMC from the name, and i'll make sure to fill up the PR template from now on Jason.

PR #101, opened 2018-01-05
cnp infrastructure

rkondratowicz:
### Change description ###
terraform files + jenkinsfile + web.config


infrastructure/variables.tf
```diff
@@ -0,0 +1,30 @@
...
+variable "location_db" {
+  type    = "string"
+  default = "West Europe"
+}
+
+variable "env" {
+  type = "string"
+}
+
+// region app config
```
lgonczar: Are these valid terraform comments?

Jenkinsfile_CNP
```diff
@@ -0,0 +1,11 @@
+#!groovy
+
+@Library("Infrastructure")
+
+def type = "java"
+def product = "draft-store"
+def app = "draft-store"
+
+withPipeline(type, product, app) {
+  enableSlackNotifications('#cmc-tech-cnp')
```
lgonczar: This should problably be `#platform-engineering`

lgonczar: Or should we have a separate channel for that?

rkondratowicz:
@njrich28 can we test this before merging?

infrastructure/main.tf
```diff
@@ -0,0 +1,33 @@
...
+  env      = "${var.env}"
+  asename  = "${data.terraform_remote_state.core_apps_compute.ase_name[0]}"
+
+  app_settings = {
+    DRAFT_STORE_DB_HOST     = "${module.db.host_name}"
+    DRAFT_STORE_DB_PORT     = "${module.db.postgresql_listen_port}"
+    DRAFT_STORE_DB_PASSWORD = "${random_string.password.result}"
+
+    IDAM_URL                = "http://idam-${var.env}.service.${data.terraform_remote_state.core_apps_compute.ase_name[0]}.internal"
+    S2S_URL                 = "http://idam-s2s-${var.env}.service.${data.terraform_remote_state.core_apps_compute.ase_name[0]}.internal"
```
rkondratowicz: for the record, none of those exist yet

infrastructure/main.tf
```diff
@@ -0,0 +1,33 @@
...
+  length = 32
+  special = true
+}
+
+module "api" {
+  source   = "git@github.com/hmcts/terraform-module-webapp.git"
+  product  = "${var.product}-api"
+  location = "${var.location_api}"
+  env      = "${var.env}"
+  asename  = "${data.terraform_remote_state.core_apps_compute.ase_name[0]}"
```
KarolJastrzebski: `basename` typo?

rkondratowicz: ase = App Service Environment

Jenkinsfile_CNP
```diff
@@ -0,0 +1,10 @@
+#!groovy
+
+@Library("Infrastructure")
+
+def type = "java"
+def product = "draft-store"
+def app = "draft-store"
```
njrich28: Internally in the pipeline, `product` and `app` get concatenated together to name the infrastructure so might be worth changing app to `service` or `backend` or something like that so the app name in the ase will be e.g. `draft-store-service-prod` rather than `draft-store-draft-store-prod`

infrastructure/main.tf
```diff
@@ -0,0 +1,33 @@
+resource "random_string" "password" {
+  length = 32
+  special = true
+}
+
+module "api" {
+  source   = "git@github.com/hmcts/terraform-module-webapp.git"
```
rkondratowicz: @njrich28 should we specify version (tag) here, or are we ok to use latest master?

njrich28: use master

infrastructure/main.tf
njrich28: that's fine - the infra will build ok and url will be exposed ok to the app - it obvs just won't work when you try to call it

PR #103, opened 2018-01-09
Sort unencrypted drafts info

rkondratowicz:
By the date it was last updated. Latest on top.


lgonczar:
Worth making sure we have a proper index for this query.

PR #105, opened 2018-03-05
Fix sonarqube warning

rkondratowicz:


src/main/java/uk/gov/hmcts/reform/draftstore/service/DraftService.java
```diff
@@ -54,14 +56,15 @@ public int create(CreateDraft newDraft, UserAndService userAndService) {
...
 
     public void update(String id, UpdateDraft updatedDraft, UserAndService userAndService) {
-        draftRepo
-            .read(toInternalId(id))
-            .map(d -> {
-                assertCanEdit(d, userAndService);
-                draftRepo.update(toInternalId(id), toDb(updatedDraft, userAndService.secrets));
-                return d;
-            })
-            .orElseThrow(() -> new NoDraftFoundException());
```
rkondratowicz: it was a dirty hack, return value was never used...

PR #118, opened 2018-04-10
Fix the syntax in Jenkinsfile_parameterized

lgonczar:
### Change description ###

Fix the syntax in Jenkinsfile_parameterized

```
[ ] Yes
[x] No
```

### DIFF
--- a/Jenkinsfile_parameterized
+++ b/Jenkinsfile_parameterized
@@ -1,6 +1,6 @@
 #!groovy
 
-@Library("Infrastructure")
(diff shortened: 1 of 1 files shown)


Jenkinsfile_parameterized
```diff
@@ -1,6 +1,6 @@
 #!groovy
 
-@Library("Infrastructure")
+@Library("Infrastructure") _
```
nitinprabhu: This need not be used as far I know.This is just for global imports

PR #119, opened 2018-04-10
Remove version from jar name and save standard output to files

lgonczar:
### Change description ###

Remove version from jar name and save standard output to files

```
[ ] Yes
[x] No
```

### DIFF
--- a/web.config
+++ b/web.config
@@ -5,6 +5,8 @@
       <add name="httpPlatformHandler" path="*" verb="*" modules="httpPlatformHandler" resourceType="Unspecified" />
     </handlers>
     <httpPlatform processPath="%JAVA_HOME%\bin\java.exe"
+                  stdoutLogEnabled="true"
+                  stdoutLogFile="%HOME%\LogFiles\stdout"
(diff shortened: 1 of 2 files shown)


web.config
```diff
@@ -5,7 +5,9 @@
       <add name="httpPlatformHandler" path="*" verb="*" modules="httpPlatformHandler" resourceType="Unspecified" />
     </handlers>
     <httpPlatform processPath="%JAVA_HOME%\bin\java.exe"
-                  arguments="-Djava.net.preferIPv4Stack=true -Dserver.port=%HTTP_PLATFORM_PORT% -jar &quot;%HOME%\site\wwwroot\draft-store-1.1.0.jar&quot;">
+                  stdoutLogEnabled="true"
+                  stdoutLogFile="%HOME%\LogFiles\stdout"
+                  arguments="-Djava.net.preferIPv4Stack=true -Dserver.port=%HTTP_PLATFORM_PORT% -jar &quot;%HOME%\site\wwwroot\draft-store.jar&quot;">
```
timja-kainos: While you're changing this I would recommend adding the app insights agent:
https://github.com/hmcts/cmc-claim-store/blob/master/web.config#L10
I'll be updating docs etc tomorrow ish

timja-kainos: also need to bundle the agent:
https://github.com/hmcts/cmc-claim-store/tree/master/lib

PR #122, opened 2018-04-12
Remove the -all suffix from jar name

lgonczar:
### Change description ###

Remove the -all suffix from jar name

```
[ ] Yes
[x] No
```

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -49,7 +49,7 @@ node {
     }
 
     stage('Package (RPM)') {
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -49,7 +49,7 @@ node {
     }
 
     stage('Package (RPM)') {
-      packager.javaRPM('draft-store', 'build/libs/draft-store-$(./gradlew -q printVersion)-all.jar',
```
rkondratowicz: is printVersion task still neded?

lgonczar: removed

PR #123, opened 2018-04-12
Enable db migrations in pipeline

lgonczar:

https://tools.hmcts.net/jira/browse/RPE-377
https://tools.hmcts.net/jira/browse/RPE-344

### Change description ###

Enable db migrations in pipeline. Also, fix the way the application talks to the DB (with current settings it can't).

```
[ ] Yes
[x] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,6 +2,7 @@ plugins {
   id 'application'
   id 'checkstyle'
   id 'io.spring.dependency-management' version '1.0.3.RELEASE'
+  id 'org.flywaydb.flyway' version '5.0.7'
   id 'org.springframework.boot' version '1.5.7.RELEASE'
   id 'org.owasp.dependencycheck' version '1.4.5.1'
   id 'com.github.ben-manes.versions' version '0.15.0'
@@ -15,6 +16,14 @@ version '1.1.0'
 
 checkstyle.toolVersion = '7.2'
 checkstyle.configFile = new File(rootDir, "checkstyle.xml")
+
+flyway {
+  url = System.getenv('FLYWAY_URL')
+  user = System.getenv('FLYWAY_USER')
+  password = System.getenv('FLYWAY_PASSWORD')
+  baselineOnMigrate = true
+  baselineVersion = '000'
+}
 
 // https://jeremylong.github.io/DependencyCheck/dependency-check-gradle/configuration.html
 dependencyCheck {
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/hmcts/reform/draftstore/config/FlywayConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.autoconfigure.flyway.FlywayAutoConfiguration;
+import org.springframework.boot.autoconfigure.flyway.FlywayMigrationStrategy;
+import org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration;
+import org.springframework.boot.autoconfigure.orm.jpa.HibernateJpaAutoConfiguration;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import uk.gov.hmcts.reform.draftstore.data.migration.FlywayNoOpStrategy;
+
+@AutoConfigureAfter({
```
lgonczar: These java classes are a copy-paste from send-letter-producer. The purpose is to prevent DB migrations from talking place at application startup.

build.gradle
```diff
@@ -78,7 +87,7 @@ dependencies {
 
   compile group: 'org.hibernate', name: 'hibernate-validator', version: '5.4.1.Final'
 
-  compile group: 'uk.gov.hmcts.reform', name: 'java-logging-spring', version: '1.4.0'
+  compile group: 'uk.gov.hmcts.reform', name: 'java-logging-spring', version: '2.2.1'
```
doncem: 😯

lgonczar: 1.4.0 didn't have the base exception referenced by one of your classes. That's why I had to bump it.

doncem: ah. then remove the base class - let's not introduce that nonsense for now

lgonczar: What about those alert levels then? Can we live without them?

doncem: in my opinion yes - we can. draft store as of this state did not have them anyway. just scope out of this PR the logging library bump and replace base class with runtime exception. then 👍 👍 

if application insights has to come in, then we have no choice but to scan all those exceptions and stick in alert levels

lgonczar: Alright then, done

PR #132, opened 2018-04-17
Enable DB migration on startup when running locally

lgonczar:
### Change description ###

Enable DB migration on startup when running the application locally

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/draftstore/config/FlywayConfiguration.java
```diff
@@ -20,7 +20,7 @@
 @AutoConfigureBefore(FlywayAutoConfiguration.class)
 @Configuration
 @ConditionalOnClass(Flyway.class)
-@ConditionalOnProperty(prefix = "flyway", name = "enabled", matchIfMissing = true)
+@ConditionalOnProperty(prefix = "dbMigration", name = "runOnStartup", havingValue = "false")
```
lgonczar: I renamed the property to avoid confusion. Flyway is never disabled - it either runs migrations or verifies them.

PR #139, opened 2018-04-25
Update dependencies

doncem:
test library update gotcha: null does not match to any

```
[ ] Yes
[x] No
```


build.gradle
```diff
@@ -72,21 +85,27 @@ sourceSets {
...
+  dependencies {
+    dependencySet(group: 'com.fasterxml.jackson.core', version: '2.9.5') {
+      entry 'jackson-annotations'
+      entry 'jackson-core'
+      entry 'jackson-databind'
+    }
+    dependencySet(group: 'com.fasterxml.jackson.datatype', version: '2.9.5') {
+      entry 'jackson-datatype-jsr310'
+    }
+  }
```
rkondratowicz: can you explain this part?

checkstyle.xml
```diff
@@ -54,13 +56,15 @@
             <property name="tokens" value="LITERAL_TRY, LITERAL_FINALLY, LITERAL_IF, LITERAL_ELSE, LITERAL_SWITCH"/>
         </module>
         <module name="NeedBraces"/>
-        <module name="LeftCurly">
-            <property name="maxLineLength" value="100"/>
```
lgonczar: Why did you remove the limit? What's the default?

doncem: I've only replaced the file with current config we have in recent project aka send-letter-service

doncem: it has been deprecated (guessing since v8) http://checkstyle.sourceforge.net/config_blocks.html#LeftCurly

build.gradle
```diff
@@ -1,20 +1,20 @@
 plugins {
   id 'application'
   id 'checkstyle'
-  id 'io.spring.dependency-management' version '1.0.3.RELEASE'
+  id 'io.spring.dependency-management' version '1.0.5.RELEASE'
   id 'org.flywaydb.flyway' version '5.0.7'
-  id 'org.springframework.boot' version '1.5.7.RELEASE'
+  id 'org.springframework.boot' version '1.5.12.RELEASE'
```
timja-kainos: spring boot 2 upgrade to come?

doncem: Yes. That's a different ball game left aside

src/test/java/uk/gov/hmcts/reform/draftstore/service/crypto/CryptoServiceTest.java
```diff
@@ -43,4 +43,30 @@ public void should_throw_an_exception_when_wrong_secret_is_used() throws Excepti
         assertThat(thrown)
             .isInstanceOf(InvalidKeyException.class);
     }
+
+    @Test
+    public void should_throw_an_exception_when_empty_input_received() {
```
lgonczar: It would be ideal to have two different tests - one for encrypting and one for decrypting. Not strong about it, so just to consider.

doncem: it would make sense if those tests actually are responsible for encrypting/decrypting. but the sole purpose is to cover the branch of conditional exception throw disregarding the following code coherence

PR #141, opened 2018-04-26
Fix preview deployment

lgonczar:
### Change description ###

Limit Vault name to 24 characters, which is the allowed maximum. Also, remove the default product name from terraform.

```
[ ] Yes
[x] No
```

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -3,9 +3,22 @@ provider "vault" {
 }
 
 locals {
-  s2s_url         = "http://rpe-service-auth-provider-${var.env}.service.${local.ase_name}.internal"
   db_connection_options      = "?ssl=true"
   ase_name                   = "${data.terraform_remote_state.core_apps_compute.ase_name[0]}"
+
+  # Environment we take our dependencies from. Typically the same as for the app,
+  # but has to be different for preview
+  dependencies_env = "${var.env == "preview"
+                          ? "aat"
+                          : var.env == "spreview" ? "saat" : var.env
+                      }"
+
+  dependencies_ase = "${var.env == "preview"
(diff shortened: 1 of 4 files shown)


infrastructure/variables.tf
```diff
@@ -1,6 +1,5 @@
 variable "product" {
   type    = "string"
-  default = "draft-store"
```
rkondratowicz: is this related?

lgonczar: Possibly. After the first commit terraform complained about not being able to find the server farm and as far as I remember that had something to do with setting the product name in code (might be wrong). This default is unnecessary, anyway.

timja-kainos: yeah it shouldn't be in here, it comes from the jenkinsfile now

lgonczar: I've changed the title of the PR to be more accurate.

infrastructure/main.tf
```diff
@@ -52,6 +52,8 @@ module "api" {
 # region save DB details to Azure Key Vault
 module "key-vault" {
   source              = "git@github.com:hmcts/moj-module-key-vault?ref=master"
+  # Making sure Vault name will not exceed the maximum length of 24 charactes
+  name                = "${substr(format("%s-%s", var.product, var.env), 0, 24)}"
```
rkondratowicz: what's the origin of this max length? can we not just increase it?

timja-kainos: netbios I believe ~30 years ago microsoft limitation ;)

lgonczar:
There's an inconsistency between Terraform state file and the actual state of the infrastructure - closing.

PR #143, opened 2018-04-26
Implement more functional tests

KarolJastrzebski:
https://tools.hmcts.net/jira/browse/RPE-342

```
[ ] Yes
[x] No
```


lgonczar:
More cases need to be covered (this pull request adds two tests). As for the tools (auth, API calls), the same code as used by smoke tests should be used here. Otherwise we'll have two versions of the same logic.

src/functionalTest/java/uk/gov/hmcts/reform/draftstore/steps/DraftStoreSteps.java
```diff
@@ -0,0 +1,89 @@
...
+            .andReturn();
+
+        DraftList draftList = response.as(DraftList.class);
+        return draftList == null ? Collections.emptyList() : draftList.data;
+    }
+
+    private RequestSpecification newRequest() {
+        String s2s = s2SAuthSteps.authenticate().assertSuccessfulAuthentication().getJwt();
+        if (StringUtils.isEmpty(s2s)) {
+            throw new IllegalArgumentException("Empty S2S token");
```
rkondratowicz: this method does not take any arguments

src/functionalTest/java/uk/gov/hmcts/reform/draftstore/steps/DraftStoreSteps.java
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        String idam = idamAuthSteps.requestRegistration().authenticate().assertSuccessfulAuthentication().getJwt();
+        if (StringUtils.isEmpty(idam)) {
+            throw new IllegalArgumentException("Empty IdAM token");
```
rkondratowicz: same

src/functionalTest/java/uk/gov/hmcts/reform/draftstore/steps/DraftStoreSteps.java
```diff
@@ -0,0 +1,89 @@
...
+        return draftList == null ? Collections.emptyList() : draftList.data;
+    }
+
+    private RequestSpecification newRequest() {
+        String s2s = s2SAuthSteps.authenticate().assertSuccessfulAuthentication().getJwt();
+        if (StringUtils.isEmpty(s2s)) {
+            throw new IllegalArgumentException("Empty S2S token");
+        }
+
+        String idam = idamAuthSteps.requestRegistration().authenticate().assertSuccessfulAuthentication().getJwt();
```
rkondratowicz: perhaps `idamJwt` instead?

src/functionalTest/java/uk/gov/hmcts/reform/draftstore/steps/DraftStoreSteps.java
```diff
@@ -0,0 +1,89 @@
...
+        return RestAssured.given()
+            .log().all()
+            .relaxedHTTPSValidation()
+            .baseUri(url)
+            .header("ServiceAuthorization", "Bearer " + s2s)
+            .header("Secret", Strings.repeat("x", 16))
+            .auth().oauth2(idam);
+    }
+
+    public Response lastResponse() {
```
rkondratowicz: do you plan to make multiple http calls using you 'steps' object? atm this method is always used after single call. can the methods making a call return a response instead?

src/functionalTest/java/uk/gov/hmcts/reform/draftstore/steps/DraftStoreSteps.java
```diff
@@ -0,0 +1,89 @@
...
+        return response;
+    }
+
+    public String createdDraftId() {
+        if (response != null) {
+            String location = response.getHeader("Location");
+            return location == null ? "" : location.substring(location.lastIndexOf("/") + 1);
+        }
+        return "";
```
rkondratowicz: `""` doesn't look like a draft Id to me

src/functionalTest/java/uk/gov/hmcts/reform/draftstore/steps/IdamAuthSteps.java
```diff
@@ -0,0 +1,125 @@
...
+        if (response != null && response.statusCode() == UNAUTHORIZED.value() && readyToRegister) {
+            newRequest()
+                .contentType(APPLICATION_JSON_VALUE)
+                .body(ImmutableMap.of(
+                    "email", email,
+                    "forename", "Platform Engineering",
+                    "surname", "Functional tests",
+                    "password", password
+                ))
+                .post("/testing-support/accounts")
```
rkondratowicz: is this endpoint enabled an all env where those test will be run on?

src/functionalTest/java/uk/gov/hmcts/reform/draftstore/steps/S2SAuthSteps.java
```diff
@@ -0,0 +1,61 @@
...
+            .contentType(APPLICATION_JSON_VALUE)
+            .body(params)
+            .when()
+            .post("/lease")
+            .andReturn();
+
+        jwt = "";
+        if (response != null && response.statusCode() == OK.value()) {
+            jwt = response.body().print();
+        }
```
rkondratowicz: looks like the test should fail right here

src/functionalTest/resources/application.properties
```diff
@@ -1 +1,10 @@
 test-url=${TEST_URL:http://localhost:8800}
+s2s-url=${S2S_URL:http://localhost:8490}
+s2s-service=${S2S_NAME:divorce}
+s2s-secret=${S2S_SECRET:AAAAA}
```
rkondratowicz: invalid default s2s secret

src/functionalTest/resources/application.properties
```diff
@@ -1 +1,10 @@
...
+idam-url=${IDAM_URL:http://localhost:8801}
```
rkondratowicz: does idam not run on 8080 by default?

src/functionalTest/java/uk/gov/hmcts/reform/draftstore/features/security/AuthenticatedRequestTest.java
```diff
@@ -29,4 +40,10 @@ public void should_reject_request_when_missing_required_headers() {
             .assertThat()
             .statusCode(400);
     }
+
+    @Test
+    public void should_accept_authenticated_service_user() {
+        assertThat(draftStoreSteps.listDrafts()).hasSize(0);
```
rkondratowicz: not related to authentication which is supposed to be tested in this class

src/functionalTest/java/uk/gov/hmcts/reform/draftstore/features/security/AuthenticatedRequestTest.java
```diff
@@ -29,4 +40,10 @@ public void should_reject_request_when_missing_required_headers() {
...
+        assertThat(draftStoreSteps.lastResponse().statusCode()).isEqualTo(OK.value());
```
rkondratowicz: Compare this test to existing test above.
it's hard to find any similarities although they're both hitting the same endpoint, the only difference being one adds jwt token headers and the other doesn't.

src/functionalTest/java/uk/gov/hmcts/reform/draftstore/features/store/CreateDraftTest.java
```diff
@@ -0,0 +1,37 @@
...
+@TestPropertySource("classpath:application.properties")
+@RunWith(SpringRunner.class)
+@ContextConfiguration(classes = FunctionalTestConfiguration.class)
+public class CreateDraftTest {
+
+    @Autowired
+    private DraftStoreSteps draftStoreSteps;
+
+    @Test
+    public void user_successfully_stores_new_draft() throws JSONException {
```
rkondratowicz: let's stick do describing behaviour, not end result

src/functionalTest/java/uk/gov/hmcts/reform/draftstore/features/store/Draft.java
```diff
@@ -0,0 +1,32 @@
...
+
+    public Draft(JSONObject json) {
+        this.json = json;
+    }
+
+    public static Draft empty() throws JSONException {
+        return create("", new JSONObject());
+    }
+
+    public static Draft create(String type, Object document) throws JSONException {
```
rkondratowicz: this method is always called with the same parameters.
Do we need this object?
Can we simply supply json string instead?

src/functionalTest/java/uk/gov/hmcts/reform/draftstore/features/store/CreateDraftTest.java
lgonczar: This name would be good for smoke tests, which are the happy scenario tests (i.e. check if we can do this). Functional tests have various edge cases and therefore should follow something like this:
```
public void should-{what we expect it do to}-when-{conditions}() { ...
```

build.gradle
```diff
@@ -153,11 +153,7 @@ dependencies {
   smokeTestCompile group: 'io.rest-assured', name: 'rest-assured', version: '3.1.0'
   smokeTestCompile group: 'com.warrenstrange', name: 'googleauth', version: '1.1.5'
 
-  functionalTestCompile sourceSets.main.runtimeClasspath
-  functionalTestCompile(
-    'org.springframework.boot:spring-boot-starter-test',
-    'io.rest-assured:rest-assured:3.1.0'
-  )
+  functionalTestCompile sourceSets.smokeTest.runtimeClasspath
```
lgonczar: Could you make sure running functional tests will not trigger smoke tests, as well?

src/functionalTest/java/uk/gov/hmcts/reform/draftstore/features/store/ReadingDraftTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+    @Test
+    public void user_can_read_own_draft() throws IOException, JSONException {
+        String documentJson = "{\"testing\":\"functional\"}";
+        String draftId = draftStoreClient.createDraft(documentJson);
+        Optional<Draft> response = draftStoreClient.readDraft(draftId);
+        assertThat(response.isPresent()).isTrue();
+        Draft draft = response.get();
+        assertThat(draft.id).isEqualTo(draftId);
+        assertThat(draft.document).isNotNull();
```
lgonczar: You could compare the whole documents, just like you do in one of the tests below.

src/functionalTest/java/uk/gov/hmcts/reform/draftstore/features/store/UpdatingDraftTest.java
```diff
@@ -0,0 +1,38 @@
...
+        String newJsonDocument = "{\"done\":true}";
+        draftStoreClient.updateDraft(draftId, newJsonDocument);
+        Optional<Draft> response = draftStoreClient.readDraft(draftId);
+
+        assertThat(response.isPresent()).isTrue();
+        Draft newDraft = response.get();
+        assertThat(newDraft.id).isEqualTo(draftId);
+        JsonNode expectedDocument = new ObjectMapper().readTree(newJsonDocument);
+        assertThat(newDraft.document).isEqualTo(expectedDocument);
+        assertThat(newDraft.created).isLessThanOrEqualTo(newDraft.updated);
```
lgonczar: Worth adding assertions checking if created and updated are not null (before this assertion). Otherwise, when we get a NullPointerException here, we won't know which one of the dates is missing.

build.gradle
KarolJastrzebski: Yep, smoke tests are not triggered.

lgonczar:
I think Nitin just meant adding `deployment_namespace` input variable

lgonczar:
Something to consider (can be changed in the next pull request) - our current naming convention for test classes is CreateTest, DeleteTest, etc. (imperative, see integration tests), so it would be best to align with that.

src/smokeTest/java/uk/gov/hmcts/reform/draftstore/response/Draft.java
```diff
@@ -26,4 +27,21 @@ public Draft(
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        Draft draft = (Draft) o;
+        return Objects.equals(id, draft.id);
```
rkondratowicz: that's not enough to conclude that the objects are equal

src/functionalTest/java/uk/gov/hmcts/reform/draftstore/features/store/RemovingDraftTest.java
```diff
@@ -0,0 +1,31 @@
...
+import uk.gov.hmcts.reform.draftstore.response.Draft;
+
+import java.io.IOException;
+import java.util.Optional;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@TestPropertySource("classpath:application.properties")
+@RunWith(SpringRunner.class)
+public class RemovingDraftTest extends SmokeTestSuite {
```
lgonczar: In one of the future pull requests it would be good to change `SmokeTestSuite` to `EndToEndTestSuite`, `AatTestSuite` or sth similar.

PR #145, opened 2018-04-26
Delete smoke test user before creation attempt

lgonczar:
### Change description ###

If there's an Idam authentication failure on user login, smoke tests make an attempt to create that user and try to log them in again. However, if the user already exists but is locked, creation won't change their state and the tests will end up in a forever failing state. That's why before creating the user we should make sure the user doesn't exist.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/smokeTest/java/uk/gov/hmcts/reform/draftstore/client/IdamClient.java
+++ b/src/smokeTest/java/uk/gov/hmcts/reform/draftstore/client/IdamClient.java
@@ -8,7 +8,9 @@ import java.util.Base64;
 import java.util.Map;
 import java.util.Optional;
 
+import static org.hamcrest.Matchers.isOneOf;
 import static org.springframework.http.HttpHeaders.CONTENT_TYPE;
+import static org.springframework.http.HttpStatus.NOT_FOUND;
 import static org.springframework.http.HttpStatus.NO_CONTENT;
 import static org.springframework.http.HttpStatus.OK;
 import static org.springframework.http.HttpStatus.UNAUTHORIZED;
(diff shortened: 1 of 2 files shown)


src/smokeTest/java/uk/gov/hmcts/reform/draftstore/client/IdamClient.java
```diff
@@ -56,6 +58,19 @@ public void registerUser() {
...
 
+    /**
+     * Deletes the user with Idam's testing support.
+     */
+    public void deleteUser() {
+        RestAssured
+            .given()
+            .relaxedHTTPSValidation()
+            .baseUri(this.idamUrl)
+            .post("/testing-support/accounts", email)
```
doncem: same post and url for deletion?

lgonczar: fixed

src/smokeTest/java/uk/gov/hmcts/reform/draftstore/client/IdamClient.java
```diff
@@ -56,6 +58,19 @@ public void registerUser() {
...
+     */
+    public void deleteUser() {
+        RestAssured
+            .given()
+            .relaxedHTTPSValidation()
+            .baseUri(this.idamUrl)
+            .delete("/testing-support/accounts/{email}", email)
+            .then()
+            .statusCode(isOneOf(NO_CONTENT.value(), NOT_FOUND.value()));
+    }
```
rkondratowicz: have you tested this with locally running idam? (given initial http method error pointed out by @doncem)

doncem: if such resource exists (`DELETE /testing-support/accounts/{email}`) i'm happy to mark this as 🚀 . no_content stands for user existence and and not_found - not found hence no deletion happened. it's fine

lgonczar: Yes, I wrote it based on what I've found in their code.

PR #146, opened 2018-04-30
Let smoke tests use Idam testing support only where it's available

lgonczar:
### Change description ###

Let smoke tests use Idam testing support only in environments where it's available (i.e. not in production)

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/smokeTest/java/uk/gov/hmcts/reform/draftstore/SmokeTestSuite.java
+++ b/src/smokeTest/java/uk/gov/hmcts/reform/draftstore/SmokeTestSuite.java
@@ -48,6 +48,9 @@ public abstract class SmokeTestSuite {
     @Value("${draft-store-url}")
     protected String draftStoreUrl;
 
+    @Value("${use-idam-testing-support")
+    protected boolean useIdamTestingSupport;
+
     protected DraftStoreClient draftStoreClient;
 
     @Before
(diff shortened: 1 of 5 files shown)


src/smokeTest/java/uk/gov/hmcts/reform/draftstore/SmokeTestSuite.java
```diff
@@ -76,7 +79,9 @@ public void tearDown() {
      * @return Idam access token
      */
     private String signIntoIdam(IdamClient idamClient) {
-        Optional<String> authorisationCode = idamClient.getAuthorisationCode(false);
+        boolean failIfUnauthorised = !useIdamTestingSupport;
+
+        Optional<String> authorisationCode = idamClient.getAuthorisationCode(failIfUnauthorised);
```
rkondratowicz: I think it would be better to have tests behave the same way on all envs if possible...

Can we stop using 'testing-support' altogether?

lgonczar: Without testing support you need to register a user manually. I've done that in production, but data in other environments is much more short-lived.

rkondratowicz: fair enough

PR #150, opened 2018-05-02
Move the database location from West Europe to UK South

lgonczar:
### Change description ###

Move the database location from West Europe to UK South.

```
[ ] Yes
[x] No
```

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -9,11 +9,14 @@ locals {
 }
 
 module "db" {
-  source              = "git@github.com:hmcts/moj-module-postgres"
-  product             = "${var.product}-db"
+  source              = "git@github.com:hmcts/moj-module-postgres?ref=cnp-449-tactical"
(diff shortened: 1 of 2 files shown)


infrastructure/main.tf
```diff
@@ -9,11 +9,14 @@ locals {
 }
 
 module "db" {
-  source              = "git@github.com:hmcts/moj-module-postgres"
-  product             = "${var.product}-db"
+  source              = "git@github.com:hmcts/moj-module-postgres?ref=cnp-449-tactical"
```
lgonczar: `.git` is unnecessary here

infrastructure/main.tf
```diff
@@ -9,11 +9,14 @@ locals {
...
+  product             = "${var.product}-postgres-db"
```
lgonczar: I had to change the name - otherwise the deployment would complain about the server already existing.

PR #152, opened 2018-05-09
Delete stale drafts with null max stale days field

rkondratowicz:
uses 'default max stale days' from config for such drafts

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/draftstore/controllers/helpers/SampleData.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/draftstore/controllers/helpers/SampleData.java
@@ -26,7 +26,7 @@ public final class SampleData {
         return new UpdateDraft("{}", null, "some_type");
     }
 
-    public static CreateDraft createDraft(int maxStaleDays) {
+    public static CreateDraft createDraft(Integer maxStaleDays) {
         return new CreateDraft(
             "{}",
             null,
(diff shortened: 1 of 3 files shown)


src/integrationTest/java/uk/gov/hmcts/reform/draftstore/controllers/helpers/SampleData.java
```diff
@@ -26,7 +26,7 @@ public static UpdateDraft updateDraft() {
         return new UpdateDraft("{}", null, "some_type");
     }
 
-    public static CreateDraft createDraft(int maxStaleDays) {
+    public static CreateDraft createDraft(Integer maxStaleDays) {
```
lgonczar: I'd probably make it an Optional, but not that important.

src/integrationTest/java/uk/gov/hmcts/reform/draftstore/data/OldDraftsCleanupTest.java
```diff
@@ -47,6 +47,25 @@ public void should_remove_stale_drafts() throws Exception {
         );
     }
 
+    @Test
+    public void should_use_default_max_stale_days_to_determine_whether_to_delete_drafts_with_null_max_stale_days() {
+        final int defaultMaxStaleDays = 30;
+        DraftStoreDao nowRepo = repoAtTime(now, defaultMaxStaleDays);
+
+        int old = nowRepo.insert("", "", SampleData.createDraft(null));
+        int ok = nowRepo.insert("", "", SampleData.createDraft(defaultMaxStaleDays + 5));
```
lgonczar: 1 would be even better than 5

src/integrationTest/java/uk/gov/hmcts/reform/draftstore/controllers/helpers/SampleData.java
rkondratowicz: constructor below takes `Integer`

PR #153, opened 2018-05-09
Set IdAM test user's password in terraform

lgonczar:
### Change description ###

This password is set only in those environments where IdAM's testing support is on. It still has to be set manually everywhere else.

```
[ ] Yes
[x] No
```

### DIFF
--- a/infrastructure/test_config.tf
+++ b/infrastructure/test_config.tf
@@ -20,3 +20,21 @@ resource "azurerm_key_vault_secret" "idam-client-secret-for-tests" {
   value     = "${data.vault_generic_secret.test_idam_client_secret.data["value"]}"
   vault_uri = "${module.key-vault.key_vault_uri}"
 }
+
+#region IdAM test user's password
+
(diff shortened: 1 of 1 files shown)


infrastructure/test_config.tf
```diff
@@ -20,3 +20,22 @@ resource "azurerm_key_vault_secret" "idam-client-secret-for-tests" {
   value     = "${data.vault_generic_secret.test_idam_client_secret.data["value"]}"
   vault_uri = "${module.key-vault.key_vault_uri}"
 }
+
+#region IdAM test user's password
+
+resource "random_string" "idam_password" {
+  length = 16
+  special = true
+  override_special = "/@\" "
```
nitinprabhu: could you please give more detail on the usage of `override_special` as this is the first I have encountered this :)

lgonczar: That was overriding the default list of special characters to use in the password. Wasn't necessary (took it from an example). Removed.

PR #154, opened 2018-05-09
Remove unencrypted document column

rkondratowicz:
Once all unencrypted drafts are deleted (by either client, or became stale) handling of this column can be removed (the new version of the service already requires all new drafts to be encrypted).

Do not merge until cmc migrates to draft store on CNP.


rkondratowicz:
Do not merge until data is migrated from tactical db

PR #159, opened 2018-05-22
Add more logging

rkondratowicz:


src/main/java/uk/gov/hmcts/reform/draftstore/service/DraftService.java
```diff
@@ -72,11 +81,13 @@ public void delete(String id, UserAndService userAndService) {
             .ifPresent(d -> {
                 assertCanEdit(d, userAndService);
                 draftRepo.delete(toInternalId(id));
+                log.info("Deleted draft. ID: {}", id);
```
lgonczar: We could also log an attempt to delete a non-existing draft. Up to you.

PR #160, opened 2018-05-23
Disable migrations on sandbox

rkondratowicz:
Currently builds on sandbox fail, as jenkins tries to run migration but does not have access to tactical env db.


Jenkinsfile_parameterized
```diff
@@ -30,6 +30,7 @@ properties([
 ])
 
 withParameterizedPipeline(params.TYPE, params.PRODUCT_NAME, params.APP, params.ENVIRONMENT, params.SUBSCRIPTION) {
-  enableDbMigration()
+  // migrations disabled temporarily, sandbox jenkins does not have access to tactical envs.
+  // enableDbMigration()
```
nitinprabhu: should we also do it for` Jenkins_CNP` file ?

PR #169, opened 2018-06-22
Update dependencies

rkondratowicz:
and suppressions

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ plugins {
   id 'io.spring.dependency-management' version '1.0.5.RELEASE'
   id 'org.flywaydb.flyway' version '5.0.7'
   id 'org.springframework.boot' version '1.5.13.RELEASE'
-  id 'org.owasp.dependencycheck' version '3.1.2'
+  id 'org.owasp.dependencycheck' version '3.2.1'
   id 'com.github.ben-manes.versions' version '0.17.0'
   id 'org.sonarqube' version '2.6.2'
   id 'jacoco'
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,7 @@ plugins {
   id 'checkstyle'
   id 'io.spring.dependency-management' version '1.0.5.RELEASE'
   id 'org.flywaydb.flyway' version '5.0.7'
-  id 'org.springframework.boot' version '1.5.10.RELEASE'
+  id 'org.springframework.boot' version '1.5.13.RELEASE'
   id 'org.owasp.dependencycheck' version '3.1.2'
   id 'com.github.ben-manes.versions' version '0.17.0'
   id 'org.sonarqube' version '2.6.2'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -103,13 +103,6 @@ dependencyManagement {
     dependencySet(group: 'com.fasterxml.jackson.datatype', version: '2.9.5') {
       entry 'jackson-datatype-jsr310'
     }
-    // the url pattern of "" is not correctly handled in tomcat
-    // it is possible for unauthorised users to gain access to web application resources that should have been protected
-    // not bumping tomcat-annotations-api as it introduces way more vulnerabilities than 8.5.27
-    dependencySet(group: 'org.apache.tomcat.embed', version: '8.5.29') {
-      entry 'tomcat-embed-core'
-      entry 'tomcat-embed-websocket'
-    }
```
rkondratowicz: no longer needed as Spring Boot update updated those as well

dependency-check-suppressions.xml
```diff
@@ -4,43 +4,20 @@
...
-    <cve>CVE-2018-1304</cve>
-    <cve>CVE-2018-1305</cve>
   </suppress>
 
   <!-- postgresql -->
 
   <suppress>
-    <notes><![CDATA[A race condition in the postgresql init script could be used by attackers able to access the postgresql account to escalate their privileges to root]]></notes>
-    <gav regex="true">org\.postgresql:postgresql:42\.1\.4</gav>
+    <notes><![CDATA[ No never version available as of yet ]]></notes>
```
doncem: no never what? 😃 

what's wrong with previous note?

dependency-check-suppressions.xml
```diff
@@ -4,43 +4,20 @@
...
-
-## CVE-2018-1305
-
-Security constraints defined by annotations of Servlets in Apache Tomcat 9.0.0.M1 to 9.0.4, 8.5.0 to 8.5.27, 8.0.0.RC1 to 8.0.49 and 7.0.0 to 7.0.84 were only applied once a Servlet had been loaded. Because security constraints defined in this way apply to the URL pattern and any URLs below that point, it was possible - depending on the order Servlets were loaded - for some security constraints not to be applied. This could have exposed resources to users who were not authorised to access them.
-]]>
-    </notes>
-    <gav regex="true">org\.apache\.tomcat:tomcat-annotations-api:8\.5\.27</gav>
+    <notes><![CDATA[ Spring Boot dependencies ]]></notes>
+    <gav regex="true">^org\.apache\.tomcat:tomcat-annotations-api:8\.5\.31*$</gav>
```
doncem: regex ends at `31`. i think it's just `\d+` until it is solved with future version

build.gradle
```diff
@@ -3,7 +3,7 @@ plugins {
   id 'checkstyle'
   id 'io.spring.dependency-management' version '1.0.5.RELEASE'
   id 'org.flywaydb.flyway' version '5.0.7'
-  id 'org.springframework.boot' version '1.5.10.RELEASE'
+  id 'org.springframework.boot' version '1.5.13.RELEASE'
```
doncem: ah. i better investigate this update. `1.5.12` was introduce heck lot's of tomcat issues

dependency-check-suppressions.xml
nitinprabhu: yeah we can keep previous note

rkondratowicz: *newer :P

rkondratowicz: @nitinprabhu the note now explain why this warning is suppressed.
(althought I personally believe it should not be suppressed and cause a build warning instead of build failure)

PR #170, opened 2018-06-22
Remove olf jenkinsfile

rkondratowicz:
Both cmc and divorce use the draf-store on CNP, no need to run builds on old env anymore.


lgonczar:
There's a typo in the commit message

PR #177, opened 2018-07-17
Changing IDAM URL to point back to tactical from strategic

ngul:
### Change description ###
Added saat.tfvars and changed IDAM url pointing to tactical Test 

```
[ ] Yes
[ X] No
```

### DIFF
--- a/infrastructure/saat.tfvars
+++ b/infrastructure/saat.tfvars
@@ -1,0 +1,1 @@
+idam_api_url = "http://betadevbccidamapplb.reform.hmcts.net/""
--- a/infrastructure/saat.tfvars
+++ b/infrastructure/saat.tfvars
@@ -1,1 +1,1 @@
-idam_api_url = "http://betadevbccidamapplb.reform.hmcts.net/""
+idam_api_url = "http://betadevbccidamapplb.reform.hmcts.net/"


infrastructure/saat.tfvars
```diff
@@ -0,0 +1 @@
+idam_api_url = "http://betadevbccidamapplb.reform.hmcts.net/""
```
timja-kainos: can we update the default one please?
CMC is happy for this to be changed till the 25th

nitinprabhu: Waiting for Divorce to confirm

Rob-Doherty: @timja-kainos + @nitinprabhu Divorce doesn't look to be using SIDAM in any env, including SAAT:
https://github.com/hmcts/div-petitioner-frontend/blob/master/infrastructure/saat.tfvars#L17
https://github.com/hmcts/div-evidence-management-client-api/blob/master/infrastructure/saat.tfvars#L2
https://github.com/hmcts/div-case-progression-service/blob/master/infrastructure/saat.tfvars#L7

Rob-Doherty: However it looks like you've got an extra `"` in there

PR #184, opened 2018-08-13
Drop port in IDAM url

timja:
Needed for SIDAM go-live, CMC have been using the version without a port for weeks,
App services don't support a port in the url, on go-live it'll be a DNS switch but the port must not be there


infrastructure/prod.tfvars
```diff
@@ -1,4 +1,4 @@
-idam_api_url = "https://prod-idamapi.reform.hmcts.net:3511"
+idam_api_url = "https://prod-idamapi.reform.hmcts.net"
```
nitinprabhu: It would be worth adding in description of PR why port is not required any more

timja: I've added more words to the description

PR #191, opened 2018-09-03
Update dependencies

rkondratowicz:


dependency-check-suppressions.xml
```diff
@@ -3,22 +3,7 @@
 
   <!-- server related, mostly applicable to linux. hopefully get better cover with spring boot 2 -->
```
doncem: can remove this line too 😉

PR #192, opened 2018-09-10
Update to new idam prod url

timja:
-

### Change description ###
Idam is migrating to the platform domain, so that after the DNS switch for SIDAM it will be on the new domain, a request should have came through for us to update it but draft-store must've been missed
https://github.com/hmcts/cmc-citizen-frontend/pull/875/files

```
[ ] Yes
[ ] No
```


timja:
@nitinprabhu err no, this change is only in prod.tfvars

PR #193, opened 2018-09-10
Populate all DB-related secrets from the output of DB module

lgonczar:

https://tools.hmcts.net/jira/browse/RPE-544

### Change description ###

Populate all DB-related secrets from the output of DB module. Before this change, there was some duplication. Also, the DB password came from Hashicorp Vault, whereas its source should be DB module creating that DB.

```
[ ] Yes
[x] No
```


lgonczar:
Of course! I should have known by the fact that the DB password wasn't being written anywhere.

PR #220, opened 2018-10-18
Enable docker build in pipeline

timja:
none

### Change description ###
Publish docker image to ACR with the cnp-java-base so that teams can use it in AKS if needed

```
[ ] Yes
[ ] No
```


run-dev.sh
```diff
@@ -1,8 +1,6 @@
 #!/usr/bin/env sh
 
-export DRAFT_STORE_DB_PASSWORD=test
-export DRAFT_STORE_DB_PORT=5432
 export S2S_USER_STUB=true
 export IDAM_USE_STUB=true
 
-./gradlew clean installDist && docker-compose build && docker-compose up
+./gradlew assemble && docker-compose build draft-store-api && docker-compose up
```
KevinJCross: should we be doing a docker-compose pull too ?

timja: nah, this is for building based on a local image, you want whatever is local.

PR #246, opened 2019-01-23
RPE-916 Automate alerts

timja:
https://tools.hmcts.net/jira/browse/RPE-916

### Change description ###
It matches what we have in production
Tested in SAAT

image

```
[ ] Yes
[ ] No
```

### DIFF
--- a/infrastructure/alert_bad_response_code.tf
+++ b/infrastructure/alert_bad_response_code.tf
@@ -1,4 +1,4 @@
-module "send-letter-up-alert" {
+module "bad-response-codes-alert" {
   source            = "git@github.com:hmcts/cnp-module-metric-alert"
   location          = "${var.location_api}"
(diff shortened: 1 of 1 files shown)


timja:
@doncem We agreed in refinement step 1 is automating the query, bringing visibility of it to use and making sure everyone gets the email, as of the current team i think only Caroline gets them.

and 
step 2 will be based on any failures looking at why its failing then improving it

I'll delete the old one once i've triggered an alert in prod with the new one

doncem:
fair enough, i thought there's going to be a next step hence the current pr completion ;)

p.s. i got the email too 👍

timja:
Cool, there's a group for the alerts in office, if you want to you can click leave group, or you can stay in it.
Bearing in mind send letter is still using the platform alerts group ^.^

PR #278, opened 2019-04-01
SIDAM in all envs (except sandbox)

markry-kainos:
### Change description ###

Update demo to SIDAM to bring in-line with AAT and Production.

```
[ ] Yes
[x] No
```


timja:
@markry-kainos I've swapped all envs except `sandbox` as it doesn't have SIDAM there
Removed some leftover rubbish too

PR #280, opened 2019-04-02
 RPE-914 Refactor AKS environment configuration in Preview

adusumillipraveen:
Refactor AKS environment configuration in Preview

https://tools.hmcts.net/jira/browse/RPE-914

### Change description ###

Refactor AKS environment configuration in Preview

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/draft-store-service/values.preview.template.yaml
+++ b/charts/draft-store-service/values.preview.template.yaml
@@ -1,0 +1,8 @@
+java:
+  environment:
+    DRAFT_STORE_DB_HOST: ${SERVICE_NAME}-postgresql  # do not modify
+    IDAM_URL: https://idam-api.aat.platform.hmcts.net
+    S2S_URL: http://rpe-service-auth-provider-aat.service.core-compute-aat.internal
+  # Don't modify below here
+  image: ${IMAGE_NAME}
+  ingressHost: ${SERVICE_FQDN}
--- a/charts/draft-store-service/values.preview.template.yaml
+++ b/charts/draft-store-service/values.preview.template.yaml
@@ -1,6 +1,4 @@
 java:
-  environment:
-    DRAFT_STORE_DB_HOST: ${SERVICE_NAME}-postgresql  # do not modify
   # Don't modify below here
   image: ${IMAGE_NAME}
   ingressHost: ${SERVICE_FQDN}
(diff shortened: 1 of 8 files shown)


charts/draft-store-service/values.preview.template.yaml
```diff
@@ -0,0 +1,8 @@
+java:
+  environment:
+    DRAFT_STORE_DB_HOST: ${SERVICE_NAME}-postgresql  # do not modify
+    IDAM_URL: https://idam-api.aat.platform.hmcts.net
```
timja: may as well put idam url and s2s url in the defaults? and not duplicate across the preview and aat files?

charts/draft-store-service/values.yaml
```diff
@@ -0,0 +1,18 @@
+java:
+  applicationPort: 8800
+  environment:
+    DRAFT_STORE_DB_HOST: "{{ .Release.Name }}-postgresql"
+    DRAFT_STORE_DB_NAME: {{ .Values.java.postgresql.postgresqlDatabase}}
+    DRAFT_STORE_DB_USER_NAME: {{ .Values.java.postgresql.postgresqlUsername}}
+    DRAFT_STORE_DB_PASSWORD: {{ .Values.java.postgresql.postgresqlDatabase}}
```
timja: how does this work if you use a secret instead?
i.e in AAT won't this still show up as an environment variable?

adusumillipraveen: It will be same as how it is today. Password will still be managed by secrets mounted onto container.

timja: if you described the pod won't it show this password though? even though its loaded from secrets

adusumillipraveen: It doesn't work somehow, will see if i can get it working. But, as far as i have read, it will just take the value from computed values.yaml . So, in this case it will display "hmcts" and not the actual password. Anyway, there won't be a postgres pod in AAT . I don't think this will be harmful . I will anyway have a look when i get this to AAT.

adusumillipraveen: It works, Are u happy for me to try and see in AAT ?

timja: yea

PR #308, opened 2019-06-13
Update owasp dependency check to version 5

luigibk:

Test for [[AB#202](https://dev.azure.com/hmcts/90472b63-dc8d-4ae6-8819-a32bd211e914/_workitems/edit/202)]

### Change description ###

```
[ ] Yes
[x] No
```

### DIFF
--- a/config/owasp/suppressions.xml
+++ b/config/owasp/suppressions.xml
@@ -14,6 +14,7 @@
   </suppress>
 
   <suppress>
+    <notes><![CDATA[Used only for testing not used in prod]]></notes>
     <gav regex="true">^org\.scala-lang:scala-(compiler|reflect|library):.*$</gav>
     <cve>CVE-2017-15288</cve>
     <cve>CVE-2010-5312</cve>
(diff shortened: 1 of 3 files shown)


config/owasp/suppressions.xml
```diff
@@ -2,14 +2,26 @@
...
+    <notes><![CDATA[Used only for testing not used in prod]]></notes>
+    <filePath regex="true">^org\.scala-lang:scala-(library|reflect|compiler):.*$</filePath>
+    <cpe>cpe:/a:scala-lang:scala</cpe>
   </suppress>
 
   <suppress>
     <notes><![CDATA[Used only for testing not used in prod]]></notes>
-    <filePath regex="true">.*postgresql-1\.10\.6\.jar</filePath>
-    <cpe>cpe:/a:postgresql:postgresql:1.10.6</cpe>
+    <gav regex="true">^org\.scala-lang:scala-(compiler|reflect|library):.*$</gav>
```
timja: could probably just suppress everything with scala? it's only used for perf testing

luigibk: In 2 minds about that as if it gets added later it's easy to miss that everything is suppressed. Even if it really should  not be added in the first place outside testing :-)

PR #313, opened 2019-06-19
Move to distroless 8 base image.

luigibk:

[[AB#250](https://dev.azure.com/hmcts/90472b63-dc8d-4ae6-8819-a32bd211e914/_workitems/edit/250)]

### Change description ###

```
[ ] Yes
[x] No
```


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM hmcts/cnp-java-base:openjdk-8u191-jre-alpine3.9-2.0.1
+FROM hmcts/cnp-java-base:openjdk-8-distroless-1.0
```
timja: don't you need to add the app insights jar if you're using distroless? as there's no curl in the image to download it at build time

luigibk: well spotted

web.config
```diff
@@ -9,7 +9,7 @@
     <httpPlatform processPath="%JAVA_HOME%\bin\java.exe"
```
timja: I think you may have problems upgrading the app insights jar by removing the old one in the same PR
I have vague memories of windows file locking problems...

You can try this approach, but I believe the way that previously worked was:

1. add new jar and point web.config to it
2. separate PR remove old jar

Note that it will likely load the old code and not the new one during the transition period because of how it loads the code.

luigibk: Mmhhh not ideal. Let's see what happens. I'll add back the previous jar if it doesn't work

PR #372, opened 2019-09-10
Update travis config

doncem:
### Change description ###

Main goal: stop wasting travis resources and block others with all the build queues aligned up on each commit which is being pushed. Period.

```
[ ] Yes
[x] No
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,34 @@
 sudo: required
+
 language: java
+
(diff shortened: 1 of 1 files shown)


doncem:
probably good bet would be transitioning to github actions instead of using travis

tbh, not sure why it took so long to apply this change. it's a common practice

PR #396, opened 2019-09-26
Adding appinsights to keyvault and bumping chart

adusumillipraveen:

```
[ ] Yes
[x] No
```


charts/draft-store-service/values.yaml
```diff
@@ -15,6 +16,5 @@ java:
   image: 'hmctspublic.azurecr.io/draft-store/service:latest'
   keyVaults:
     "draft-store":
-      resourceGroup: draft-store-service
       secrets:
         - service-POSTGRES-PASS
```
timja: you do need to add the app insights key here?
AppInsightsInstrumentationKey

adusumillipraveen: will do in next PR :)

PR #702, opened 2020-09-08
Upgrade terraform provider

luigibk:

Upgrade Terraform Azure provider + Terraform version.

Upgrade key-vault module.

### Change description ###

```
[ ] Yes
[ ] No
```


infrastructure/main.tf
```diff
@@ -1,5 +1,16 @@
-provider "azurerm" {
-  version = "1.44.0"
+terraform {
```
timja, line 1: might cause a problem you already have this defined in state.tf

https://github.com/hmcts/draft-store/blob/master/infrastructure/state.tf#L1

luigibk: 👍

timja (commented):
Missing adding .terraform-version file, and I would probably update the terraform syntax also.

`terraform 0.12upgrade` should do it for you

Otherwise IntelliJ can do it as well

timja (approved):

PR #705, opened 2020-09-09
Fix quality gate error (as sonar likes?)

luigibk:
### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/draftstore/data/DraftStoreDao.java
+++ b/src/main/java/uk/gov/hmcts/reform/draftstore/data/DraftStoreDao.java
@@ -63,10 +63,10 @@ public class DraftStoreDao {
             keyHolder,
             new String[] {"id"}
         );
-        if (keyHolder.getKey() == null) {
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/reform/draftstore/data/DraftStoreDao.java
```diff
@@ -63,10 +63,10 @@ public int insert(String userId, String service, CreateDraft newDraft) {
             keyHolder,
             new String[] {"id"}
         );
-        if (keyHolder.getKey() == null) {
-            throw new DataRetrievalFailureException("Unable to retrieve the generated key [null] or key not generated.");
+        if (keyHolder.getKey() != null) {
+            return keyHolder.getKey().intValue();
         }
```
timja, line 68: I think you might still get pinged maybe, perhaps something like?
```suggestion
		var key = keyHolder.getKey()
        if (key != null) {
            return key.intValue();
        }
```

timja (approved):

PR #706, opened 2020-09-09
Fix quality gate error (sonar driven syntax)

luigibk:
### Change description ###

```
[ ] Yes
[ ] No
```


timja (approved):
ha not using my `var` syntax suggestion? :ton

luigibk:
> ha not using my `var` syntax suggestion? :ton

I am not a big fan of local variable inference, but it's a matter of taste really

PR #836, opened 2021-03-16
Spring boot 2.4 upgrade with configtree

adusumillipraveen:
- Use spring boot configtree instead of properties-volume-springboot-starter
- Change to use `disableKeyvaults`in preview as `SPRING_CLOUD_PROPERTIESVOLUME_ENABLED` won't work anymore.
- Using alias in secretproviderclass to rename secrets to env vars/spring properties.
- Also upgrades other dependencies as there were bunch of vulnerabilities.

```
[ ] Yes
[x] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -194,7 +194,7 @@ configurations.all {
         details.useVersion '2.12.10'
       }
       if (details.requested.group in ['org.apache.tomcat.embed']) {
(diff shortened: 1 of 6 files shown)


luigibk (approved):

timja (approved):
nice simple one

PR #918, opened 2021-11-15
Add index on user_id and service

timja:
CMC said they were moving away from draft store partly because of performance.

I checked prod and the only performance issue appears to be this missing index =/


taomoh (approved):

njrich28 (approved):
LGTM

PR #1224, opened 2023-02-14
DTSPO-11969 - Add expiresAfter in sbox

endakelly:
https://tools.hmcts.net/jira/browse/DTSPO-11969

### Change description ###
Add expiresAfter tag to resources in sbox
Updates to metric alert module will cause the existing template resource to be deleted and recreated due to switching from a deprecated resource to a new version

```
[ ] Yes
[x] No
```


thomast1906 (approved):
Not long now to this expiry

PR #1343, opened 2023-06-02
DTSPO-13832 - Add depends_on for app insights

endakelly:
https://tools.hmcts.net/jira/browse/DTSPO-13832

### Change description ###
Fixing error related to app insights being in region disallowed by policy

```
[ ] Yes
[x] No
```

### DIFF
--- a/infrastructure/alert_bad_response_code.tf
+++ b/infrastructure/alert_bad_response_code.tf
@@ -22,4 +22,6 @@ EOF
   trigger_threshold          = "0"
   resourcegroup_name         = azurerm_resource_group.rg.name
   common_tags                = var.common_tags
+
+  depends_on = [ azurerm_application_insights.appinsights ]
 }
--- a/infrastructure/alert_bad_response_code.tf
+++ b/infrastructure/alert_bad_response_code.tf
@@ -23,5 +23,4 @@ EOF
   resourcegroup_name         = azurerm_resource_group.rg.name
   common_tags                = var.common_tags
 
-  depends_on = [ azurerm_application_insights.appinsights ]
 }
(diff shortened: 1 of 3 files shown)


timja (changes requested):
update line 4 instead to use an output from the resource

explicit dependencies should only be used in very limited situations
https://developer.hashicorp.com/terraform/tutorials/configuration-language/dependencies#manage-explicit-dependencies

endakelly:
> update line 4 instead to use an output from the resource
> 
> explicit dependencies should only be used in very limited situations [developer.hashicorp.com/terraform/tutorials/configuration-language/dependencies#manage-explicit-dependencies](https://developer.hashicorp.com/terraform/tutorials/configuration-language/dependencies#manage-explicit-dependencies)

That was my initial thought as well but the resource doesn't export a name atrtibute https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/application_insights#attributes-reference

The app insights instance isn't showing up as to be created nor is it in the state which is odd. Do you think this lifecycle block could be the cause and is it still needed? https://github.com/hmcts/draft-store/blob/121251a870c6f79453d456e930ecbb7f5e14f579/infrastructure/main.tf#L184-L189

Could also just not create any app insights related resources in sbox.

endakelly:
Ignore my last, actually just noticed there is an error about being denied by policy due to the location being West Europe. Seems like it was set to that a few years ago, maybe when UK South wasn't live. Would it be best to add an exemption to allow app insights to exist in West Europe or move the existing ones to UK South? I'd imagine that would delete the existing insights data and I don't see a move option in the portal.

timja:
> Ignore my last, actually just noticed there is an error about being denied by policy due to the location being West Europe. Seems like it was set to that a few years ago, maybe when UK South wasn't live. Would it be best to add an exemption to allow app insights to exist in West Europe or move the existing ones to UK South? I'd imagine that would delete the existing insights data and I don't see a move option in the portal.

just re-create in uk south i think

timja (approved):

PR #1524, opened 2024-07-16
DTSPO-17910 - Change create_mode

endakelly:
https://tools.hmcts.net/jira/browse/DTSPO-17912

### Change description ###
Change create mode to default from update
This is causing a 404 error when deploying but the update mode should only be used when updating the postgres version as well
Screenshot 2024-07-16 at 15 00 54

```
[ ] Yes
[x] No
```


cpareek (commented):
oh I can see its just Sandbox, never mind
Are we just testing to see what happens at the time of upgrade?

cpareek (approved):

